PR #168, opened 2016-05-22
Identical checker

aurpine:
Adds new identical checker that returns true only when output is **exactly** the same as the answer.


Xyene:
Thanks!

PR #177, opened 2016-06-04
Add sysinfo to PY2

WallE256:
During self-testing on Ubuntu 16.04, Python 2 would get a segmentation fault. This is the change to fix this.


Xyene:
Looks good, thanks!

PR #224, opened 2016-11-26
Update RUST.py

mmun:
- Add rand crate
- Loosen dmoj crate version
- Don't try to be clever with the package name. It was possible for the rust compiler to fail if the problem name had non-identifier characters.


mmun:
@quantum5 Thanks. I've fixed that and rust self-tests are passing now.

PR #233, opened 2017-01-03
Include pre-generated Cargo.lock for Rust

mmun:
This allows cargo to bypass hitting the network to determine which dependencies to use.


Xyene:
Travis failed for issue unrelated to this PR.

PR #244, opened 2017-03-10
Create Interactive Grader

BinAccel:
Writing your own interactor can be painful, and this should remove most of the pain.

### DIFF
--- a/dmoj/graders/__init__.py
+++ b/dmoj/graders/__init__.py
@@ -1,3 +1,4 @@
 from dmoj.graders.standard import StandardGrader
 from dmoj.graders.signature import SignatureGrader
 from dmoj.graders.custom import CustomGrader
+from dmoj.graders.interactive import IneractiveGrader, Interactor
--- a/dmoj/graders/__init__.py
+++ b/dmoj/graders/__init__.py
@@ -1,4 +1,4 @@
 from dmoj.graders.standard import StandardGrader
 from dmoj.graders.signature import SignatureGrader
 from dmoj.graders.custom import CustomGrader
-from dmoj.graders.interactive import InteractiveGrader, Interactor
+from dmoj.graders.interactive import InteractiveGrader
--- a/dmoj/graders/__init__.py
+++ b/dmoj/graders/__init__.py
@@ -1,4 +1,4 @@
 from dmoj.graders.standard import StandardGrader
 from dmoj.graders.signature import SignatureGrader
 from dmoj.graders.custom import CustomGrader
-from dmoj.graders.interactive import IneractiveGrader, Interactor
+from dmoj.graders.interactive import InteractiveGrader, Interactor
--- a/dmoj/graders/interactive.py
+++ b/dmoj/graders/interactive.py
@@ -3,7 +3,7 @@ from dmoj.graders.standard import StandardGrader
 class Interactor(object):
     def __init__(self, process):
         self.process = process
-        self.closed = False
+        self._is_closed = False
 
     def read(self):
         return self.process.stdout.read()
@@ -16,13 +16,13 @@ class Interactor(object):
         self.process.stdin.flush()
 
     def writeln(self, str):
-        write(str + '\n')
+        self.write(str + '\n')
 
     def close(self):
-        if not closed:
+        if not self._is_closed:
             for stream in [self.process.stdin, self.process.stdout, self.process.stderr]:
                 stream.close()
-            closed = True
+            self._is_closed = True
 
 class InteractiveGrader(StandardGrader):
     def _interact_with_process(self, case, result, input):
--- a/dmoj/graders/interactive.py
+++ b/dmoj/graders/interactive.py
@@ -45,12 +45,12 @@ class Interactor(object):
             raise WrongAnswer
         return ret
 
-    def write(self, str):
-        self.process.stdin.write('%s' % str)
+    def write(self, val):
+        self.process.stdin.write(str(val))
         self.process.stdin.flush()
 
-    def writeln(self, str):
-        self.write('%s\n' % str)
+    def writeln(self, val):
+        self.write(str(val) + '\n')
 
     def close(self):
         for stream in [self.process.stdin, self.process.stdout, self.process.stderr]:
@@ -61,7 +61,7 @@ class InteractiveGrader(StandardGrader):
         interactor = Interactor(self._current_proc)
         self.check = False
         try:
-            self.check = self.interact(interactor)
+            self.check = self.interact(case, interactor)
             interactor.close()
         except (WrongAnswer, IOError):
             pass
--- a/dmoj/graders/interactive.py
+++ b/dmoj/graders/interactive.py
@@ -1,9 +1,13 @@
 from dmoj.graders.standard import StandardGrader
 
+class WrongAnswer(BaseException):
+    pass
+
 class Interactor(object):
     def __init__(self, process):
         self.process = process
         self._is_closed = False
+        self._tokens = None
 
     def read(self):
         return self.process.stdout.read()
@@ -11,6 +15,33 @@ class Interactor(object):
     def readln(self):
         return self.process.stdout.readline()
 
+    def readtoken(self, str=None):
+        if not self._tokens:
+            self._tokens = self.readln()
+        try:
+            ret, self._tokens = self._tokens.split(str, 1)
+        except ValueError:
+            ret = self._tokens
+            self._tokens = None
+        return ret
+
+    def readint(self, lo=float('-inf'), hi=float('inf'), str=None):
+        try:
+            ret = int(self.readtoken(str))
+        except ValueError:
+            raise WrongAnswer
+        if lo > ret or ret > hi:
+            raise WrongAnswer
+        return ret
+
+    def readfloat(self, lo=float('-inf'), hi=float('inf'), str=None):
+        try:
+            ret = float(self.readtoken(str))
+        except ValueError:
+            raise WrongAnswer
+        if lo > ret or ret > hi:
+            raise WrongAnswer
+        return ret
     def write(self, str):
         self.process.stdin.write(str)
         self.process.stdin.flush()
@@ -31,7 +62,7 @@ class InteractiveGrader(StandardGrader):
         try:
             self.check = self.interact(interactor)
             interactor.close()
-        except:
+        except WrongAnswer:
             pass
 
         self._current_proc.wait()
@@ -39,3 +70,5 @@ class InteractiveGrader(StandardGrader):
     def check_result(self, case, result):
         return self.check
 
+    def interact(self, interactor):
+        raise NotImplementedError
--- a/dmoj/graders/interactive.py
+++ b/dmoj/graders/interactive.py
@@ -12,9 +12,10 @@ class Interactor(object):
         return self.process.stdout.read()
 
     def readln(self, strip_newline=True):
+        ret = self.process.stdout.readline()
         if strip_newline:
-            return self.process.stdout.readline().rstrip()
-        return self.process.stdout.readline()
+            return ret.rstrip()
+        return ret
 
     def readtoken(self, delim=None):
         if not self._tokens:
--- a/dmoj/graders/interactive.py
+++ b/dmoj/graders/interactive.py
@@ -15,33 +15,34 @@ class Interactor(object):
     def readln(self):
         return self.process.stdout.readline()
 
-    def readtoken(self, str=None):
+    def readtoken(self, delim=None):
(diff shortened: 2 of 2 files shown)


dmoj/graders/interactive.py
```diff
@@ -0,0 +1,40 @@
...
+            interactor.close()
+        except NotImplementedError as e:
+            # it is not the user's fault that the problemsetter was incompetent
+            raise e
+        except:
+            self.check = False
+
+        self.process.wait()
+        
+    def interact():
```
Xyene: This isn't the same signature as what's used in L28.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,40 @@
+from dmoj.graders.standard import StandardGrader
+
+class Interactor(object):
+    def __init__(self, process):
+        self.process = process
+
+    def read(self):
+        return self.process.stdout.read()
+
+    def readline(self):
```
Xyene: `readline` and `writeln`: one of these things is not like the other.

And it's `readline`.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,40 @@
...
+        return self.process.stdout.readline()
+
+    def write(self, str):
+        self.process.stdin.write(str)
+        self.process.stdin.flush()
+        
+    def writeln(self, str):
+        write(str + '\n')
+        
+    def close(self):
```
Xyene: `close`-ing a second time will cause an error &mdash; better have a boolean check here.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,40 @@
...
+    def _interact_with_process(self, case, result, input):
+        interactor = Interactor(self._current_proc)
+        try:
+            self.check = self.interact(interactor)
+            interactor.close()
+        except NotImplementedError as e:
+            # it is not the user's fault that the problemsetter was incompetent
+            raise e
+        except:
+            self.check = False
```
Xyene: Only a `WrongAnswer` exception should set `check=False`. Everything else should propagate into an IE, otherwise this could get _very_ confusing.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,40 @@
+from dmoj.graders.standard import StandardGrader
+
+class Interactor(object):
```
Xyene: This class should also implement something among the lines of `readint`, `readtoken` etc., and catch `ValueError`s as necessary, rethrowing them as a new `WrongAnswer` exception (see below comments).

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def write(self, str):
+        self.process.stdin.write(str)
+        self.process.stdin.flush()
+
+    def writeln(self, str):
+        write(str + '\n')
+
+    def close(self):
+        if not closed:
```
Xyene: "What are instance variables?"

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,41 @@
+from dmoj.graders.standard import StandardGrader
+
+class Interactor(object):
+    def __init__(self, process):
+        self.process = process
+        self.closed = False
```
Xyene: `_is_closed`  is a better name seeing as it is both a boolean and internal.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,41 @@
...
+            self.check = self.interact(interactor)
+            interactor.close()
+        except:
+            pass
+
+        self._current_proc.wait()
+
+    def check_result(self, case, result):
+        return self.check
+
```
Xyene: Where did the `interact` stub go?? :'(

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,41 @@
...
+    def readln(self):
+        return self.process.stdout.readline()
+
+    def write(self, str):
+        self.process.stdin.write(str)
+        self.process.stdin.flush()
+
+    def writeln(self, str):
+        write(str + '\n')
```
quantum5: In case @Xyene's comment is not obvious.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def close(self):
+        if not closed:
+            for stream in [self.process.stdin, self.process.stdout, self.process.stderr]:
+                stream.close()
+            closed = True
```
quantum5: Yet again.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,41 @@
...
+
+class InteractiveGrader(StandardGrader):
+    def _interact_with_process(self, case, result, input):
+        interactor = Interactor(self._current_proc)
+        self.check = False
+        try:
+            self.check = self.interact(interactor)
+            interactor.close()
+        except:
+            pass
```
quantum5: You need correct error handling here.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,74 @@
...
+        if lo > ret or ret > hi:
+            raise WrongAnswer
+        return ret
+
+    def readfloat(self, lo=float('-inf'), hi=float('inf'), str=None):
+        try:
+            ret = float(self.readtoken(str))
+        except ValueError:
+            raise WrongAnswer
+        if lo > ret or ret > hi:
```
Xyene: `if not lo < ret < hi`

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,74 @@
...
+        self._is_closed = False
+        self._tokens = None
+
+    def read(self):
+        return self.process.stdout.read()
+
+    def readln(self):
+        return self.process.stdout.readline()
+
+    def readtoken(self, str=None):
```
Xyene: `delim` is a better name than `str`

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def readfloat(self, lo=float('-inf'), hi=float('inf'), str=None):
+        try:
+            ret = float(self.readtoken(str))
+        except ValueError:
+            raise WrongAnswer
+        if lo > ret or ret > hi:
+            raise WrongAnswer
+        return ret
+    def write(self, str):
```
Xyene: Missing newline above

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,75 @@
...
+        self.process.stdin.write(str)
+        self.process.stdin.flush()
+
+    def writeln(self, str):
+        self.write(str + '\n')
+
+    def close(self):
+        if not self._is_closed:
+            for stream in [self.process.stdin, self.process.stdout, self.process.stderr]:
+                stream.close()
```
Xyene: It is possible for this to `IOError`; don't want that to propagate up to IE

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, process):
+        self.process = process
+        self._is_closed = False
+        self._tokens = None
+
+    def read(self):
+        return self.process.stdout.read()
+
+    def readln(self):
+        return self.process.stdout.readline()
```
Xyene: I may be wrong, but I think this will return a newline that most people probably don't care about. `rstrip` would come in handy.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def readtoken(self, delim=None):
+        if not self._tokens:
+            self._tokens = self.readln()
+        try:
+            ret, self._tokens = self._tokens.split(delim, 1)
+        except ValueError:
```
Xyene: When can this `raise ValueError`?

BinAccel: `ret, self._tokens = 'abc'.split(' ', 1)`

Xyene: Ah, because it can't unpack. OK.

quantum5: Why not just split it when read and pop it off one by one?

BinAccel: What if they want different delims each time?

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,75 @@
...
+            ret = self._tokens
+            self._tokens = None
+        return ret
+
+    def readint(self, lo=float('-inf'), hi=float('inf'), delim=None):
+        try:
+            ret = int(self.readtoken(delim))
+        except ValueError:
+            raise WrongAnswer
+        if not lo < ret < hi:
```
quantum5: Is open interval really the best way?

BinAccel: [That's what Tudor said](https://github.com/DMOJ/judge/pull/244#discussion_r105577740)

Xyene: If only because I didn't read your code fully; closed interval is indeed the best option here.

quantum5: He is telling you how to express yourself better, not whether the behaviour is good. @Xyene

dmoj/graders/interactive.py
quantum5: Fine.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,77 @@
...
+class Interactor(object):
+    def __init__(self, process):
+        self.process = process
+        self._tokens = None
+
+    def read(self):
+        return self.process.stdout.read()
+
+    def readln(self, strip_newline=True):
+        if strip_newline:
```
Xyene: This code duplication is awful, please store the `readline` into a variable and then optionally `rstrip` on it.

dmoj/graders/__init__.py
```diff
@@ -1,3 +1,4 @@
 from dmoj.graders.standard import StandardGrader
 from dmoj.graders.signature import SignatureGrader
 from dmoj.graders.custom import CustomGrader
+from dmoj.graders.interactive import InteractiveGrader, Interactor
```
quantum5: I fail to see how anyone would ever benefit from being able to do `from dmoj.graders import Interactor`.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,78 @@
...
+        try:
+            ret = float(self.readtoken(delim))
+        except ValueError:
+            raise WrongAnswer
+        if not lo <= ret <= hi:
+            raise WrongAnswer
+        return ret
+
+    def write(self, str):
+        self.process.stdin.write(str)
```
Xyene: This can also `IOError`.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,78 @@
...
+class Interactor(object):
+    def __init__(self, process):
+        self.process = process
+        self._tokens = None
+
+    def read(self):
+        return self.process.stdout.read()
+
+    def readln(self, strip_newline=True):
+        ret = self.process.stdout.readline()
```
Xyene: This can `IOError`.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,78 @@
...
+class WrongAnswer(BaseException):
+    pass
+
+class Interactor(object):
+    def __init__(self, process):
+        self.process = process
+        self._tokens = None
+
+    def read(self):
+        return self.process.stdout.read()
```
Xyene: This too.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,78 @@
...
+class InteractiveGrader(StandardGrader):
+    def _interact_with_process(self, case, result, input):
+        interactor = Interactor(self._current_proc)
+        self.check = False
+        try:
+            self.check = self.interact(interactor)
+            interactor.close()
+        except WrongAnswer:
+            pass
+
```
Xyene: `IOErrors` should collectively be handled here.

dmoj/graders/interactive.py
```diff
@@ -0,0 +1,75 @@
...
+            stream.close()
+
+class InteractiveGrader(StandardGrader):
+    def _interact_with_process(self, case, result, input):
+        interactor = Interactor(self._current_proc)
+        self.check = False
+        try:
+            self.check = self.interact(interactor)
+            interactor.close()
+        except WrongAnswer, IOError:
```
Xyene: This does not do what you think it does.

quantum5: Don't be bad. This means `except` an exception of class `WrongAnswer` and bind it to the name `IOError`.

Xyene: (you need brackets if you want it to be interpreted like that)

PR #251, opened 2017-04-30
Don't print problem directories

BinAccel:
Forgot to remove this line


Xyene:
Already fixed sometime or other.

PR #260, opened 2017-05-17
schema for init.yml

Yukarinn:


dmoj/problem.py
```diff
@@ -33,6 +34,8 @@ def __init__(self, problem_id, time_limit, memory_limit, load_pretests_only=Fals
             doc = yaml.safe_load(self.problem_data['init.yml'])
             if not doc:
                 raise InvalidInitException('I find your lack of content disturbing.')
+            init_check = Core(source_file=doc, schema_files=["judge/dmoj/problem_schema.yml"])
```
quantum5: Why don't you do `os.path.join(os.path.dirname(__file__), 'problem_schema.yml)` like a normal person? This only works when run from a certain directory...

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,142 @@
...
+      "entry":
+        type: str
+      "header":
+        type: str
+  "custom_judge":
+    type: str
+  "io_redirects":
+    type: map
+    mapping:
+      "DATA.in":
```
Xyene: The key can actually be any filename.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,142 @@
+#normal names
+type: map
+mapping:
+  "archive":
```
Xyene: No need for quotes; this goes for a bunch of other places too.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,142 @@
+#normal names
```
Xyene: `# Normal cases`

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,142 @@
...
+
+#defaults
+  "wall_time_factor":
+    type: int
+  "output_prefix_length":
+    type: int
+  "output_limit_length":
+    type: int
+
+#names with ++
```
Xyene: This is awful but I can't think of a better way to do it, so just put it at the end of the file to conceal the awfulness.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,142 @@
...
+# outer points, in, out
+  "points":
+    type: int
+  "in":
+    type: str
+  "out":
+    type: str
+
+# pretest_test_cases
+  "pretest_test_cases":
```
Xyene: This actually has the exact same format as `test_cases`. You could probably group them, maybe.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,142 @@
...
+  "custom_judge":
+    type: str
+  "io_redirects":
+    type: map
+    mapping:
+      "DATA.in":
+        type: map
+        mapping:
+          "fd":
+            type: str
```
Xyene: `fd` is not necessarily a `str`. The options for it are:

* `stdout`
* `stdin`
* `stderr`
* any integer

Xyene: Though if this is hard to do in pykwalify, I'd leave it as is.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,142 @@
...
+    matching: "any"
+    sequence:
+# normal cases
+      - type: map
+        mapping:
+          "points":
+            type: int
+          "in":
+            type: str
+          "out":
```
Xyene: Missing optional `generator_args` seq of `str`, both here and in batched cases.

dmoj/problem_schema.yml
Xyene: Missing `unbuffered` bool node.

dmoj/problem_schema.yml
Xyene: Can trim down some of these because some really don't make sense in practice. Only `test_cases`, `pretest_test_cases`, and `batched` make sense in practice. A unit test will start failing after this, but you can edit it to not be stupid.

/cc @quantum5

quantum5: inb4 `output_prefix_length+: __import__('random').randint(1, 100)` in some April Fools problem, though whoever does that really should be punished severely.

dmoj/cli.py
```diff
@@ -330,23 +331,19 @@ def execute(self):
         print highlight(file_diff, guess_lexer(file_diff), terminal256())
```
Xyene: Should move [`_find_exe`](https://github.com/DMOJ/judge/blob/e922b9f7867958c53b2de97942bbc628d49629ff/dmoj/cptbox/sandbox.py#L22) into a helper in utils, then check if `_find_exe('git.exe' if os.name == 'nt' else 'git')` exists (function doesn't raise OSError), and if so use `subprocess.call(_find_exe(...), 'diff', 'file1', 'file2')` instead.

dmoj/cli.py
```diff
@@ -316,9 +316,10 @@ def execute(self, line):
             print ansi_style('#ansi[%s](yellow)/#ansi[%s](green) in %s' % (problem, i + 1, lang))
         print
         
+
 class DifferenceCommand(Command):
-    name: 'diff'
-    help: 'Shows difference between two files.'
+    name = 'diff'
+    help = 'Shows difference between two files.'
```
Xyene: Should also allow showing difference between two submissions, or a mix of submissions and files (if argument is an integer, it can be assumed to be a submission).

If `git` exists in `PATH`, selected submissions would be written to a temp file [`tempfile`](https://docs.python.org/2/library/tempfile.html) module before being passed to `git diff`.

dmoj/cli.py
```diff
@@ -330,23 +331,19 @@ def execute(self):
         print highlight(file_diff, guess_lexer(file_diff), terminal256())
 
 class ShowSubmissonIdOrFilename(Command):
-    name: 'show'
-    help: 'Shows file based on submission ID or filename'
+    name = 'show'
+    help = 'Shows file based on submission ID or filename'
 
     def _populate_parser(self):
         self.arg_parser.add_argument('submission_id', type=int, help='id of submission to show')
```
Xyene: Should have only one parameter, `id_or_source`, and assume it's a submission id if it's an `int`, source file if it's not.

dmoj/cli.py
```diff
@@ -330,23 +331,19 @@ def execute(self):
         print highlight(file_diff, guess_lexer(file_diff), terminal256())
 
 class ShowSubmissonIdOrFilename(Command):
-    name: 'show'
-    help: 'Shows file based on submission ID or filename'
+    name = 'show'
+    help = 'Shows file based on submission ID or filename'
```
Xyene: Missing `.` at end

dmoj/cli.py
BinAccel: What about Windows?

Xyene:
Here's a mostly functional implementation of the `diff` command:

```python
class DifferenceCommand(Command):
    name = 'diff'
    help = 'Shows difference between two files.'

    def _populate_parser(self):
        self.arg_parser.add_argument('id_or_source_1', help='id or path of first source', metavar='<source 1>')
        self.arg_parser.add_argument('id_or_source_2', help='id or path of second source', metavar='<source 2>')

    def get_data(self, id_or_source):
        err = None
        src = None
        try:
            id = int(id_or_source)
            try:
                global graded_submissions
                problem, lang, src, tl, ml = graded_submissions[id - 1]
            except IndexError:
                err = "invalid submission '%d'" % (id - 1)
        except ValueError:
            try:
                with open(os.path.realpath(id_or_source), 'r') as f:
                    src = f.read()
            except Exception as io:
                 err = str(io)

        return src, err

    def execute(self, line):
        args = self.arg_parser.parse_args(line)

        data1, err1 = self.get_data(args.id_or_source_1)
        data2, err2 = self.get_data(args.id_or_source_2)

        file_diff = '\n'.join(list(difflib.unified_diff(data1.splitlines(), data2.splitlines(), fromfile=args.id_or_source_1, tofile=args.id_or_source_2, lineterm='')))
        print pygments.highlight(file_diff, pygments.lexers.DiffLexer(), pygments.formatters.Terminal256Formatter())
```

quantum5:
Since when are we doing kitchen sink level pull requests? I thought all features are supposed to be separate?

dmoj/cli.py
```diff
@@ -310,6 +317,81 @@ def execute(self, line):
...
+        err = None
+        src = None
+        lexer = None
+        try:
+            id = int(id_or_source)
+            try:
+                global graded_submissions
+                problem, lang, src, tl, ml = graded_submissions[id - 1]
+                if lang in ['PY2', 'PYPY2', 'PY3', 'PYPY3']:
+                    lexer = pygments.lexers.PythonLexer()
```
quantum5: There is a python 3 lexer.

dmoj/problem.py
```diff
@@ -33,6 +35,10 @@ def __init__(self, problem_id, time_limit, memory_limit, load_pretests_only=Fals
             doc = yaml.safe_load(self.problem_data['init.yml'])
             if not doc:
                 raise InvalidInitException('I find your lack of content disturbing.')
+            logging.disable(logging.CRITICAL)  # disable logging
```
quantum5: No you disable that specific logger.

quantum5:
This should have been superseded by the other two pull requests.

PR #263, opened 2017-06-05
Yukarinn schema

Yukarinn:


dmoj/problem.py
```diff
@@ -33,6 +35,10 @@ def __init__(self, problem_id, time_limit, memory_limit, load_pretests_only=Fals
             doc = yaml.safe_load(self.problem_data['init.yml'])
             if not doc:
                 raise InvalidInitException('I find your lack of content disturbing.')
+            logging.disable(logging.CRITICAL)  # disable logging for pykwalify
```
quantum5: Get rid of this ridiculous global solution for a local problem.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,155 @@
+#normal names
```
quantum5: Always use a space after `#`. Don't be so inconsistent too.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,155 @@
...
+          mode:
+            type: str
+  hints:
+    type: seq
+    sequence:
+      - type: str
+  unbuffered:
+    type: bool
+
+#defaults
```
quantum5: Always use a space after `#`. Don't be so inconsistent too.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,155 @@
...
+          in:
+            type: str
+          out:
+            type: str
+          generator_args:
+            type: seq
+            sequence:
+              - type: str
+
+#normal test_cases
```
quantum5: Always use a space after `#`. Don't be so inconsistent too.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,155 @@
...
+                  in:
+                    type: str
+                  out:
+                    type: str
+                  generator_args:
+                    type: seq
+                    sequence:
+                      - type: str
+
+#test_cases ++
```
quantum5: Always use a space after `#`. Don't be so inconsistent too.

dmoj/problem_schema.yml
```diff
@@ -0,0 +1,155 @@
...
+  test_cases+:
+    type: str
+  test_cases++:
+    type: str
+#names with ++
```
quantum5: Always use a space after `#`. Don't be so inconsistent too.

dmoj/problem.py
```diff
@@ -33,6 +35,9 @@ def __init__(self, problem_id, time_limit, memory_limit, load_pretests_only=Fals
             doc = yaml.safe_load(self.problem_data['init.yml'])
             if not doc:
                 raise InvalidInitException('I find your lack of content disturbing.')
+            logging.getLogger("logging").setLevel(logging.CRITICAL)
```
Xyene: This should be called on `pykwalify.core` instead of the global `logging`, so that not _all_ logs are silenced. It'd also probably be good to not have it be set every time a problem is loaded (just once is enough), so calling it somewhere like [here](https://github.com/DMOJ/judge/blob/master/dmoj/judge.py#L584) should be good.

I think that's about the last thing left with this PR.

PR #266, opened 2017-06-19
Generalize line comparison of linecount.py

Phoenix1369:
- Move extraneous output check to the beginning
- Handle pretests correctly

### DIFF
--- a/dmoj/checkers/linecount.py
+++ b/dmoj/checkers/linecount.py
@@ -18,7 +18,7 @@ def check(process_output, judge_output, point_value, feedback=False, **kwargs):
     match = equal
 
     # Overload lambda - passed as a string literal
-    if 'match' in kwargs and isinstance(kwargs['match'], basestring):
+    if 'match' in kwargs:
         try:
             # Skips assignment if eval fails
             match = eval(kwargs['match'])
--- a/dmoj/checkers/linecount.py
+++ b/dmoj/checkers/linecount.py
@@ -3,7 +3,7 @@ from dmoj.result import CheckerResult
 verdict = u"\u2717\u2713"
 
 def check(process_output, judge_output, point_value,
-        feedback=False, match=lambda x,y: x.strip()==y.strip(), **kwargs):
+        feedback=False, match=lambda p, j: p.strip()==j.strip(), **kwargs):
 
     process_lines = filter(None, process_output.strip().split("\n"))
     judge_lines = filter(None, judge_output.strip().split("\n"))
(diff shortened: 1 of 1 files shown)


dmoj/checkers/linecount.py
```diff
@@ -3,7 +3,7 @@
 verdict = u"\u2717\u2713"
 
 def check(process_output, judge_output, point_value,
-        feedback=False, match=lambda p, j: p.strip()==j.strip(), **kwargs):
+        feedback=False, match=lambda p, j: p.strip() == j.strip(), **kwargs):
```
BinAccel: Add 2 more spaces please

dmoj/checkers/linecount.py
```diff
@@ -1,19 +1,40 @@
...
     judge_lines = filter(None, judge_output.strip().split("\n"))
 
+    # Extraneous output
+    if len(process_lines) > len(judge_lines):
+        return False
+
+    match = equal
+
+    # Overload lambda - passed as a string literal
+    if 'match' in kwargs:
```
quantum5: I think you should just check `isinstance` string instead of doing this madness.

dmoj/checkers/linecount.py
```diff
@@ -1,19 +1,40 @@
...
+        try:
+            # Skips assignment if eval fails
+            match = eval(kwargs['match'])
+        except Exception as e:
+            print(e.__doc__, e.message, file=sys.stderr, sep='\n')
```
quantum5: How does `e.__doc__` even become helpful?

dmoj/checkers/linecount.py
```diff
@@ -1,19 +1,40 @@
+from __future__ import print_function
```
quantum5: I don't see why this is necessary.

dmoj/checkers/linecount.py
quantum5: In any case, don't catch the exception and just let it propagate into internal error, which this is, instead of silently eating it.

dmoj/checkers/linecount.py
```diff
@@ -2,18 +2,28 @@
...
     judge_lines = filter(None, judge_output.strip().split("\n"))
 
+    if len(process_lines) > len(judge_lines):
+        return False
+
+    if not judge_lines:
+        return True
+
+    # Overload lambda - passed as a string literal
+    if isinstance(match, str):
```
quantum5: What about unicode?

BinAccel: I remember him saying that's why he used `basestring`

Issue #212, opened 2016-08-22
Batch scoring functions


Xyene:
Some IOI problems have partial batch scoring: things like minimum score of batch cases, maximum score, score while factoring in submission size, etc. We should support a couple of these common cases, and allow a `CustomGrader` to implement its own.

quantum5:
This requires site-side change, since batches just numbers attached to `SubmissionTestCase` objects.

PR #270, opened 2017-09-01
Allow batches to not short circuit

kiritofeng:
Allow for a key, `sc`, within the `init.yml` to dictate on a per-batch basis whether or not said batch should be short circuited. Can be combined with #212 .

### DIFF
--- a/dmoj/problem.py
+++ b/dmoj/problem.py
@@ -38,6 +38,7 @@ class Problem(object):
                 'output_prefix_length': 64,
                 'output_limit_length': 25165824,
                 'binary_data': False,
+                'short_circuit': True,
             })
         except (IOError, ParserError, ScannerError) as e:
             raise InvalidInitException(str(e))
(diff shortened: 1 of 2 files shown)


dmoj/problem.py
```diff
@@ -99,7 +99,9 @@ def __del__(self):
 
 class BatchedTestCase(object):
     def __init__(self, batch_no, config, problem):
-        self.config = config
+        self.config = ConfigNode(config.raw_config, defaults={
```
quantum5: What?? What about the things inherited from the global level??? Why don't you move this to the global config?

dmoj/problem.py
```diff
@@ -99,9 +100,10 @@ def __del__(self):
 
 class BatchedTestCase(object):
     def __init__(self, batch_no, config, problem):
-        self.config = ConfigNode(config.raw_config, defaults={
-            'short_circuit': True
-        })
+#        self.config = ConfigNode(config.raw_config, defaults={
```
quantum5: Dude WTF is this comment?

PR #274, opened 2017-09-25
Prevent GO from throwing IE if package is not main

kiritofeng:


dmoj/executors/GO.py
```diff
@@ -32,3 +36,9 @@ def get_version_flags(cls, command):
 
     def get_nproc(self):
         return [-1, 1][os.name == 'nt']
+
+    def create_files(self, problem_id, source_code, *args, **kwargs):
+        super(Executor, self).create_files(problem_id, source_code, *args, **kwargs)
```
quantum5: Why even create the file physically?

PR #281, opened 2017-11-26
Fixes for Coffee, PHP, and R

WillyPillow:
On my Debian Stretch box, the three languages need these modifications to pass self-tests. However, I have to admit that I'm not well-versed in security stuff, and it would be great if someone can double-check that this patch does not create security implications :)

### DIFF
--- a/dmoj/executors/COFFEE.py
+++ b/dmoj/executors/COFFEE.py
@@ -7,7 +7,7 @@ class Executor(ScriptExecutor):
     ext = '.coffee'
     name = 'COFFEE'
     nproc = -1
-    fs = ['/etc/(?:resolv|nsswitch).conf$', '/$']
+    fs = ['/etc/(?:resolv|nsswitch).conf$', '/$', '/dev/pts/.*$']
     command = 'node'
     syscalls = ['newselect', 'select', 'pipe2', 'poll', 'write', 'epoll_create1',
                 'eventfd2', 'epoll_ctl', 'epoll_wait', 'sched_yield', 'restart_syscall',
--- a/dmoj/executors/R.py
+++ b/dmoj/executors/R.py
@@ -16,9 +16,9 @@ class Executor(ScriptDirectoryMixin, ScriptExecutor):
 
     if os.name != 'nt':
         syscalls = ['mkdir', 'setup', 'fork', 'waitpid', 'wait4', 'getpgrp', 'execve',
-                    ('statfs64', ACCESS_DENIED), ('statfs', ACCESS_DENIED)]
+                    ('statfs64', ACCESS_DENIED), ('statfs', ACCESS_DENIED), 'newfstatat', 'unlinkat', 'openat']
 
-    fs = ['/etc/passwd$', '/etc/nsswitch.conf$', '/etc/group$']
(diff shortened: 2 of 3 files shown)


outlawmose:
I'm sorry I have no idea how I got subscribed. I have no idea or knowledge of this subject.

dmoj/executors/COFFEE.py
```diff
@@ -7,7 +7,7 @@ class Executor(ScriptExecutor):
     ext = '.coffee'
     name = 'COFFEE'
     nproc = -1
-    fs = ['/etc/(?:resolv|nsswitch).conf$', '/$']
+    fs = ['/etc/(?:resolv|nsswitch).conf$', '/$', '/dev/pts/.*$']
```
Xyene: This is suboptimal, since it'd mean a judged process can read the commands of any user connected to a terminal. We do suggest people run the judge in its own user, which would raise EPERM when CoffeeScript tries accessing that. You can try replicating that like we used to in the SBCL executor:

https://github.com/DMOJ/judge/blob/a145b4f8a31a0e0c1232eb89458e8b8917256ac4/dmoj/executors/SBCL.py#L26-L43

(of course, `readlink` may not be the appropriate syscall here)

dmoj/executors/R.py
```diff
@@ -16,9 +16,9 @@ class Executor(ScriptDirectoryMixin, ScriptExecutor):
 
     if os.name != 'nt':
         syscalls = ['mkdir', 'setup', 'fork', 'waitpid', 'wait4', 'getpgrp', 'execve',
-                    ('statfs64', ACCESS_DENIED), ('statfs', ACCESS_DENIED)]
+                    ('statfs64', ACCESS_DENIED), ('statfs', ACCESS_DENIED), 'newfstatat', 'unlinkat', 'openat']
```
Xyene: This is definitely unsafe, it'd let a judged process open and unlink any file the user it is running under has access to (this would, for example, include problem data or the judge code itself).

R is not really feasible to run reliably in a `ptrace` sandbox, unfortunately. It updates frequently, always breaking in new creative ways. On the other hand, it hasn't been used so solve more than "Hello, World!" on dmoj.ca in 2 years, so it likely doesn't matter.

Xyene:
@WillyPillow these would be better as separate PRs; the PHP one looks fine, the other two need some work to be secure.

@outlawmose I believe you watched the repository sometime, you can unwatch it [here](https://github.com/DMOJ/judge/subscription) to stop receiving notifications.

WillyPillow:
@Xyene  Thanks for your suggestions! However, seeing that you've made some (seemingly related) changes in #282, is this PR still relevant?

Xyene:
@WillyPillow #282 is not entirely related to this PR, it was more of a general cleanup of allowed syscalls that was long overdue. Directly impacting this PR is that we've removed the R executor, which never worked satisfactorily in the first place.

For PHP, your best bet is to use the `extra_fs` option in your judge config:

```yaml
extra_fs:
    PHP: [".*/conf.d$"]
```

PHP installs vary far too much by distribution in config file placement to have a "catch-all" for them, so it's best left up to individual operators.

The CoffeeScript issue is something we might be able to resolve judge-side, particularly if the runtime continues to run even if access to those files is blocked as per my previous comment.

WillyPillow:
Sorry for the delay (was a bit busy recently).

So that leaves Coffee. However, I just realized that, without the workaround, even though it throws a sys_open error while self-testing, submissions work just fine. Thus, I guess this PR isn't really needed :P

PR #302, opened 2017-12-30
cptbox improvement package

quantum5:
This originally started as the arm64 porting effort, but has wide-ranging consequences on `cptbox`.

The main changes are:

* Improved ARM support
* ARM64 support
* ARM-on-ARM64 support
* Use `process_vm_read` to read whole pages instead of `long`s with `ptrace(PTRACE_PEEKDATA)` when possible
* Efforts to make cptbox compilable under clang/bionic on Android
* Correct handling of modern *at system calls
* Add restrictions to previously unrestricted system calls like `stat`, `readlink`, and their *at forms
* A new `pre_syscall`/`post_syscall` approach to read all registers instead of summoning `ptrace` to do the dirty work for every edit. This is currently only used on ARM64, but I suspect will improve ptbox on x86 and ARM as well. `strace` uses this approach unconditionally.
* Improved logging on CI


quantum5:
Our codebase will be greatly simplified if we used `PTRACE_GETREGSET` on x86, since we can then remove the x86-on-x64 debugger and stick to a single debugger for 32-bit. However, `PTRACE_GETREGSET` is introduced in Linux 2.6.34 which is in neither the RHEL 5 or RHEL 6 kernel used by OpenVZ, and so will break compatibility with the original target platform.

@Xyene what do you think?

Xyene:
During self-test of Ruby 2 on ARM64:

```
ptrace(PTRACE_GETREGSET): No such process
INFO:dmoj.cptbox:Protection fault on: 73 (sys_ppoll)
INFO:dmoj.cptbox:Arg0: 0x00004000006b09e0
INFO:dmoj.cptbox:Arg1: 0x0000000000000002
INFO:dmoj.cptbox:Arg2: 0x0000000000000000
INFO:dmoj.cptbox:Arg3: 0x0000000000000000
INFO:dmoj.cptbox:Arg4: 0x0000000000000000
INFO:dmoj.cptbox:Arg5: 0x0000000000000000
```

dmoj/cptbox/ptdebug.cpp
```diff
@@ -84,9 +102,67 @@ void pt_debugger::poke_reg(int idx, long data) {
 typedef int ptrace_read_t;
 #else
 typedef long ptrace_read_t;
+
+#ifndef SYS_process_vm_readv
+#define SYS_process_vm_readv 270
+#endif
+
+ssize_t __attribute__((weak)) process_vm_readv(
```
Xyene: Since this is [supposedly](http://man7.org/linux/man-pages/man2/process_vm_readv.2.html) in `sys/uio.h`, a comment as to why/what platform doesn't have it would be useful.

dmoj/cptbox/ptdebug.cpp
```diff
@@ -84,9 +102,67 @@ void pt_debugger::poke_reg(int idx, long data) {
...
+    return syscall(SYS_process_vm_readv, (long) pid, lvec, liovcnt, rvec, riovcnt, flags);
+}
 #endif
 
 char *pt_debugger::readstr(unsigned long addr, size_t max_size) {
+#if PTBOX_FREEBSD
+    return readstr_peekdata(addr, max_size);
+#else
+    static unsigned long page_size = -sysconf(_SC_PAGESIZE);
+    static unsigned long page_mask = (unsigned long) page_size;
```
Xyene: Doesn't this mean `page_mask == page_size`, or is something sketchy going on here

dmoj/cptbox/ptdebug_arm.cpp
```diff
@@ -26,7 +26,11 @@ int pt_debugger_arm::syscall() {
 }
 
 void pt_debugger_arm::syscall(int id) {
+#ifdef PTRACE_SET_SYSCALL
+    ptrace(PTRACE_SET_SYSCALL, tid, 0, id);
+#else
     poke_reg(ARM_r7, id);
```
Xyene: "¬`PTRACE_SET_SYSCALL` ⇒ compiling on ARM" is a pretty convoluted assumption, I'd reverse it to `#ifdef PTBOX_ARM` or something like that instead.

dmoj/cptbox/ptdebug_arm64.cpp
```diff
@@ -0,0 +1,120 @@
...
+void pt_debugger_arm64::result(long value) {
+    poke_reg(ARM_x0, value);
+}
+
+long pt_debugger_arm64::arg0() {
+    return peek_reg(ARM_x0);
+}
+
+void pt_debugger_arm64::arg0(long data) {
+#if !PTBOX_FREEBSD
```
Xyene: DMOJ Judge on ARM64 FreeBSD?? 😲

dmoj/cptbox/ptproc.cpp
```diff
@@ -232,21 +235,28 @@ int pt_process::monitor() {
...
                                 exit_reason = protection_fault(syscall);
                                 continue;
                             default:
                                 // Default is to kill, safety first.
-                                //printf("Killed by DISALLOW or None: %d\n", syscall);
+                                // printf("Killed by DISALLOW or None: %d\n", syscall);
                                 exit_reason = protection_fault(syscall);
                                 continue;
                         }
+                    } else if (!callback(context, syscall)) {
```
Xyene: Might want a comment here that this is specifically for out-of-range syscalls on ARM platforms, probably linking https://github.com/DMOJ/judge/issues/128 for good measure.

dmoj/cptbox/ptdebug_arm64.cpp
```diff
@@ -0,0 +1,120 @@
...
+    struct iovec iovec;
+    iovec.iov_base = arm64_reg;
+    iovec.iov_len = sizeof arm64_reg;
+    if (ptrace(PTRACE_SETREGSET, tid, NT_PRSTATUS, &iovec))
+        perror("ptrace(PTRACE_SETREGSET)");
+}
+#endif
+
+long pt_debugger_arm64::peek_reg(int reg) {
+    return arm64_reg[reg];
```
Xyene: The way this is set up makes it seem like `arm64_reg[NT_ARM_SYSTEM_CALL]` will always return the most recently-set syscall number, but that behaviour depends on how `syscall(int id)` was compiled. Don't know if that's of any value, though, but it might prevent some weird issue in the future.

dmoj/cptbox/sandbox.py
```diff
@@ -214,15 +221,21 @@ def r_execution_time(self):
...
         except OSError:
             import traceback
             traceback.print_exc()
 
     def _callback(self, syscall):
-        callback = self._callbacks[syscall]
+        try:
+            callback = self._callbacks[syscall]
+        except IndexError:
+            if self._syscall_index == 3:
```
Xyene: Do the ARM-specific syscalls not exist on ARM64?

dmoj/executors/GASARM.py
```diff
@@ -11,7 +11,7 @@ class Executor(GASExecutor):
     platform_prefixes = ['arm-linux-gnueabihf']
 
     qemu_path = env.runtime.qemu_arm
-    dynamic_linker = env.runtime['ld.so_x86'] or '/lib/ld-linux-armhf.so.3'
+    dynamic_linker = env.runtime['ld.so_arm'] or '/lib/ld-linux-armhf.so.3'
     crt_pre = env.runtime.crt_pre_x86 or ['/usr/lib/arm-linux-gnueabihf/crt1.o', '/usr/lib/arm-linux-gnueabihf/crti.o']
```
Xyene: Is `crt_pre_x86` what's needed on ARM?

dmoj/cptbox/chroot.py
```diff
@@ -258,5 +263,7 @@ def do_tgkill(self, debugger):
         return tgid == debugger.pid
 
     def do_prctl(self, debugger):
+        # PR_GET_DUMPABLE = 3
         # PR_SET_NAME = 15
-        return debugger.arg0 in (15,)
+        # PR_SET_VMA = 0x53564d41, used on Android
```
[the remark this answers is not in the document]
quantum5: I mean, the ARM systems we will encounter mostly will be either Android or Raspberry Pi... so it better work on both.

dmoj/cptbox/helper.cpp
```diff
@@ -62,9 +62,9 @@ int cptbox_child_run(const struct child_config *config) {
     setrlimit2(RLIMIT_STACK, RLIM_INFINITY);
     setrlimit2(RLIMIT_CORE, 0);
 
-    if (config->stdin >= 0)  dup2(config->stdin, 0);
-    if (config->stdout >= 0) dup2(config->stdout, 1);
-    if (config->stderr >= 0) dup2(config->stderr, 2);
+    if (config->stdin_ >= 0)  dup2(config->stdin_, 0);
+    if (config->stdout_ >= 0) dup2(config->stdout_, 1);
+    if (config->stderr_ >= 0) dup2(config->stderr_, 2);
```
quantum5: In case you are wondering, some compilers, like *cough* clang *cough*, define `std*` to be macros because some brain damaged clause in the C standard said they need to be macros.

dmoj/cptbox/ptbox.h
```diff
@@ -18,6 +19,10 @@
 #   define PTBOX_FREEBSD 0
 #endif
 
+#if defined(__arm64__) || defined(__aarch64__)
+#   define PTBOX_NEED_PRE_POST_SYSCALL
```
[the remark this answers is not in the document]
quantum5: Eventually, `PTBOX_NEED_PRE_POST_SYSCALL` will be used on every platform. Using one system call to take the entire register block out of the debugged process seems to be the most efficient solution, even on x86.

dmoj/cptbox/ptdebug.cpp
quantum5: Indeed. Android's bionic C doesn't have the wrapper defined even though the underlying kernel has the system call.

dmoj/cptbox/ptdebug.cpp
quantum5: Oh shit this is a bug.

dmoj/cptbox/ptdebug_arm.cpp
quantum5: Our current setup requires compiling ARM stuff on x86, so we need some silly hacks.

dmoj/cptbox/ptdebug_arm64.cpp
```diff
@@ -0,0 +1,120 @@
...
+#define NT_ARM_SYSTEM_CALL 0x404
+#endif
+    struct iovec iovec;
+    iovec.iov_base = &id;
+    iovec.iov_len = sizeof id;
+
+    if (ptrace(PTRACE_SETREGSET, tid, NT_ARM_SYSTEM_CALL, &iovec))
+        perror("ptrace(PTRACE_SETREGSET, NT_ARM_SYSTEM_CALL)");
+#else
+    poke_reg(ARM_x8, id);
```
quantum5: Note that this will never work, but it's needed to compile on x86.

dmoj/cptbox/ptdebug_arm64.cpp
quantum5: We cannot change the system call register since it's x8, but the system call overwrites x0. On x86, the system call number register is the one trashed by the return value. This is not the case on ARM.

dmoj/cptbox/ptdebug_arm64.cpp
quantum5: It just breaks the build on FreeBSD.

dmoj/cptbox/ptproc.cpp
quantum5: It's not just ARM, but it's there to catch other stupidity.

dmoj/cptbox/sandbox.py
quantum5: Definitely not. The ARM system call table is not the same as the ARM64 one. The ARM64 is the new unified table with the minimum number of system calls.

dmoj/executors/GASARM.py
quantum5: No.

dmoj/cptbox/ptbox.h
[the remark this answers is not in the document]
Xyene: RIP OpenVZ 2014-2018

PR #304, opened 2017-12-31
Normalize `process_output`

kiritofeng:
Normalize `process_output`, which will prevent checkers like `floats`, etc, from failing on Windows due to the presence of `\r\n`.


quantum5:
I think this might be ridiculous. A probably better solution is to use `re.split` in the checkers, since this actually doubles the memory requirement when grading large submissions using the standard checker, which is completely tolerant.

@Xyene what do you think?

PR #307, opened 2018-01-11
Attempt to untoast Ruby CI test

kiritofeng:
Adding `/home` to Ruby's `extra_fs` allows the auto-configuration to work.


quantum5:
I said to add the escaped and `$`'d version every component of `ruby_home` to `fs`. Letting ruby reading basically any user file on travis is not a productive way of fixing this. Might as well disable the sandbox and have it "pass" self-test that way. :vomit:

PR #319, opened 2018-03-31
Add a dmoj-rest server

simonvpe:
This patch adds a `dmoj-rest` server. It's an initial draft that's good enough for my use case but it definitely needs improvements. Please leave feedback!


maradanasai:
Hi @simonvpe I'm actually looking for this. One question is that we should specify `memoryLimit` in KB's or MB's?

Issue #311, opened 2018-02-24
Class-finding regex doesn't match Unicode classnames


Xyene:
https://dmoj.ca/submission/800804, Rövarspråket

PR #325, opened 2018-07-24
Detect Unicode Classes in Java; #311

kiritofeng:
Modified the regexes in `dmoj.executor.java_executor` to account for Unicode submissions, and added a unit test for malformed UTF-8.

### DIFF
--- a/dmoj/executors/java_executor.py
+++ b/dmoj/executors/java_executor.py
@@ -204,7 +204,7 @@ class JavacExecutor(JavaExecutor):
         # This step is necessary because of Unicode classnames
         try:
             source_code = utf8text(source_code)
-        except UnicodeEncodeError:
+        except UnicodeDecodeError:
             raise CompileError('Your UTF-8 is bad, and you should feel bad')
         class_name = find_class(source_code)
         self._code = self._file('%s.java' % class_name.group(1))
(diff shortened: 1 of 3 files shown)


dmoj/executors/java_executor.py
```diff
@@ -13,13 +13,11 @@
...
-restring = re.compile(br''''(?:\\.|[^'\\])'|"(?:\\.|[^"\\])*"''', re.DOTALL)
-reinline_comment = re.compile(br'//.*?(?=[\r\n])')
-reclass = re.compile(br'\bpublic\s+(?:strictfp\s+)?(?:(?:abstract|final)\s+)?(?:strictfp\s+)?class\s+([_a-zA-Z\$][_0-9a-zA-z\$]*?)\b')
-repackage = re.compile(br'\bpackage\s+([^.;]+(?:\.[^.;]+)*?);')
-redeunicode = re.compile(br'\\u([0-9a-f]{4})', re.I)
-deunicode = lambda x: redeunicode.sub(lambda a: six.unichr(int(a.group(1), 16)), x)
+recomment = re.compile(r'/\*.*?\*/', re.DOTALL)
+restring = re.compile(r''''(?:\\.|[^'\\])'|"(?:\\.|[^"\\])*"''', re.DOTALL)
+reinline_comment = re.compile(r'//.*?(?=[\r\n])')
+reclass = re.compile(r'\bpublic\s+(?:strictfp\s+)?(?:(?:abstract|final)\s+)?(?:strictfp\s+)?class\s+([\w\$][\w\$]*?)\b',re.U)
```
quantum5: Need a space before `re.U`. You should probably make all of them `re.U` for consistently, since on python 3, that's the default, and we don't want python 2 to run differently.

dmoj/tests/test_malformed_unicode.py
```diff
@@ -0,0 +1,22 @@
+import os
+import unittest
+
+from dmoj.error import CompileError
+from dmoj.executors.JAVA8 import Executor as JAVA8Executor
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
```
quantum5: Why do you need mock???

PR #327, opened 2018-07-26
Fix Appveyor CI Errors

kiritofeng:
Now passes all tests except for the `VC` executor, which is not required, and the `java8_unicode_classname` test, which is a Windows/Java issue.

### DIFF
--- a/.appveyor.test.py
+++ b/.appveyor.test.py
@@ -4,8 +4,6 @@ import logging
 import os
 import sys
 
-import codecs
-
 from dmoj import judgeenv, executors
 from dmoj.testsuite import Tester
 from dmoj.utils.ansi import ansi_style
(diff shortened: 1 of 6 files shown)


dmoj/utils/unicode.py
```diff
@@ -22,9 +22,13 @@ def utf8text(maybe_bytes, errors='strict'):
 
 
 def unicode_stdout_stderr():
+    sys.stdout = codecs.getwriter('utf-8')(sys.stdout)
+    sys.stderr = codecs.getwriter('utf-8')(sys.stderr)
+'''
```
quantum5: Undo this, and do not call `unicode_stdout_stderr` in `.appveyor.test.py`.

.appveyor.test.py
```diff
@@ -14,11 +14,11 @@
 
 
 def main():
-    unicode_stdout_stderr()
     logging.basicConfig(level=logging.INFO)
 
     sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
     sys.stderr = os.fdopen(sys.stderr.fileno(), 'w', 0)
+    unicode_stdout_stderr()
```
quantum5: Inline what you currently have for `unicode_stdout_stderr` since you are going to break it for linux like this.

dmoj/executors/base_executor.py
```diff
@@ -344,7 +345,9 @@ def get_compile_process(self):
                   'preexec_fn': self.create_executable_limits(), 'time_limit': self.compiler_time_limit}
         kwargs.update(self.get_compile_popen_kwargs())
 
-        return self.TimedPopen(self.get_compile_args(), **kwargs)
+        args = self.get_compile_args()
+
+        return self.TimedPopen(args, **kwargs)
```
quantum5: Just put this back to whatever it was...

PR #336, opened 2018-09-27
Fixups for running DMOJ on Android

Xyene:


Xyene:
It ran on Termux pre-@kiritofeng's changes, so we should be good. I don't think the sandbox worked well (I think everything passed?), but I think this is good enough for now to not outright crash.

PR #342, opened 2018-10-11
add identical_with_pe checker

xiaowuc1:
This adds an augmented checker that raises a presentation error as feedback to the user in the event that the standard DMOJ checker would have accepted the given output, but the output does not exactly match that expected of the judge.


dmoj/checkers/identical_with_pe.py
```diff
@@ -0,0 +1,13 @@
+from dmoj.checkers._checker import standard
+from dmoj.result import CheckerResult
+from dmoj.utils.unicode import utf8bytes
+
+def check(process_output, judge_output, **kwargs):
+    # return AC iff the output matches exactly
```
Xyene: Implementation-wise this checker looks good to me, but I feel this should be integrated into the regular `identical` checker, and have PEs be opt-out rather than opt-in via a checker argument (see the `floats` checker for how `init.yml`-configured `precision` is passed in).

Issue #242, opened 2017-02-28
Cached executor framework


Xyene:
We should provide some sort of API to cache executors for a defined period of time and postpone cleanup until then. The use-case are compiled executors needed by checkers (offloading heavy computation to C++) or generators.

Compilation for some executors (for example, those using testlib.h) is particularly expensive. In a contest scenario where the same problem is expected to be submitted frequently, recompiling the executor is both inefficient and prohibitively slow (testlib takes up to 20 seconds to compile!)

BinAccel:
The same should be allowed for custom checkers for similar reasons.

Xyene:
As long as an API exists, checkers can use it with no extra work.

quantum5:
How about `get_cached_executor(class, *args, **kwargs)` that looks up the stuff in an LRU cache? Parameters can be decided by command line, or `judge.yml`.

Xyene:
That'd be the way to go. Right now, the generator does something similar in terms of caching one generator for the entire submission.

Also from the looks of it, we don't currently free up the generator (we rely on `__del__` being called, eventually).

PR #347, opened 2018-10-18
Cache compiled generator for duration of judge run

Xyene:
Would encourage the usage of generators more frequently, when they don't need to be recompiled each run.

### DIFF
--- a/dmoj/problem.py
+++ b/dmoj/problem.py
@@ -1,6 +1,7 @@
 import os
 import subprocess
 import zipfile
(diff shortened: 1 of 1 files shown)


quantum5:
Might be wise to pull in something like https://pypi.org/project/pylru/ or https://pypi.org/project/lru-dict/ so we don't cache too much. I feel uneasy about this growing indefinitely.

Xyene:
Superseded by #386, a full implementation of #242.

PR #360, opened 2018-12-07
Add unittest capability

outloudvi:


outloudvi:
Wrong base selected. Sorry :)

Issue #364, opened 2018-12-08
Problem resource linking support


Xyene:
## Motivation
It is currently very hard to manage the addition problems that need access to different resources than standard problems. An example of a problem that is difficult to support as-is is any that requires the presence of certain libraries that would trivialize other problems on the judge (e.g. `scipy`, `numpy`).

## Proposed Solution
A simple solution for this would be to allow the automatic symlinking of some problem YAML-defined files or directories into each submission's temporary directory. Read rights would be inherited, while writes would continue to be disallowed.

This plays particularly well with #363; it is easy to imagine a `datasci` template that provides symlinks to `numpy` and `scipy`.

## Proposed Configuration
In `init.yml`:

```yaml
symlinks: {
    numpy: /code/runtimes/python/numpy,
    scipy: /code/runtimes/python/scipy
}
```

## Supersedes `fd_redirects`
After this feature is implemented, the current `fd_redirects` can be much more intuitively implemented using symlinks.

```yaml
symlinks: {
    DATA1.in: /dev/stdin,
    DATA1.out: /dev/stdout
}
```

Where multiple cases are present and symlinking manually becomes a pain, dynamic keys can easily be used.

Riolku:
Curious: now that we resolve symlinks... does this work?

PR #367, opened 2018-12-11
Implement support for symlinking problem resources; #364

Xyene:


dmoj/executors/mixins.py
```diff
@@ -114,6 +114,19 @@ def get_env(self):
                 return env
 
             def launch(self, *args, **kwargs):
+                if 'symlinks' in kwargs:
+                    for src, dst in kwargs['symlinks'].items():
```
quantum5: Why not `for src, dst in kwargs.get('symlinks', {}).items()`?

dmoj/executors/mixins.py
```diff
@@ -114,6 +114,19 @@ def get_env(self):
...
+                        src = os.path.abspath(os.path.join(self._dir, src))
+                        # Disallow the creation of symlinks outside the submission directory.
+                        if os.path.commonprefix([src, self._dir]) == self._dir:
+                            # If a link already exists under this name, it's probably from a
+                            # previous case, but might point to something different.
+                            if os.path.islink(src):
+                                os.unlink(src)
+                            os.symlink(dst, src)
+                        else:
+                            raise InternalError('cannot symlink outside of submission directory')
```
quantum5: Did you import `InternalError`?

dmoj/executors/mixins.py
Xyene: I think this code is more explicit in what it's actually doing. Note that on all *our* invocations of `launch`, symlinks will be an empty dict due to the default config entry.

quantum5: I fail to see how the fact it's always on our invocations makes any difference. I just think it fits better with the rest of the file to use `.get` instead of `in` and `__getitem__`, as well as reduce the amount of indentation.

dmoj/executors/mixins.py
```diff
@@ -114,6 +115,18 @@ def get_env(self):
...
+                    src = os.path.abspath(os.path.join(self._dir, src))
+                    # Disallow the creation of symlinks outside the submission directory.
+                    if os.path.commonprefix([src, self._dir]) == self._dir:
+                        # If a link already exists under this name, it's probably from a
+                        # previous case, but might point to something different.
+                        if os.path.islink(src):
+                            os.unlink(src)
+                        os.symlink(dst, src)
+                    else:
+                        raise InternalError('cannot symlink outside of submission directory')
```
quantum5: This sounds like the destination cannot be outside the submission directory.

PR #372, opened 2018-12-13
Remove IO redirects in favour of IO symlinks

Xyene:
**NOT TO BE MERGED** until all problems on dmoj.ca are updated to not rely on `io_redirects`.

### DIFF
--- a/dmoj/graders/standard.py
+++ b/dmoj/graders/standard.py
@@ -32,10 +32,11 @@ class StandardGrader(BaseGrader):
 
         input = case.input_data()  # cache generator data
 
-        self._current_proc = self.binary.launch(time=self.problem.time_limit, memory=self.problem.memory_limit,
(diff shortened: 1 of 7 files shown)


dmoj/graders/standard.py
```diff
@@ -32,8 +32,9 @@ def grade(self, case):
 
         input = case.input_data()  # cache generator data
 
-        self._current_proc = self.binary.launch(time=self.problem.time_limit, memory=self.problem.memory_limit,
-                                                pipe_stderr=True, io_redirects=case.io_redirects(),
+        self._current_proc = self.binary.launch(time=self.problem.time_limit,
+                                                memory=self.problem.memory_limit,
+                                                pipe_stderr=True,
```
quantum5: This one short line bothers me more than it should.

PR #381, opened 2018-12-25
Implement seccomp filtering optimization

Xyene:
This change has been tested to work on x64 for a couple of critical
executors. There might exist some loose ends on ARM.

`seccomp` is used to remove the need to return from `ptrace` for all
syscalls that occur, allowing us to selectively filter more dangerous
syscalls (like `open`), while incurring a smaller penalty for frequent,
but harmless, syscalls (like `write`).

There are a number of semantic differences between pure `ptrace` and
`ptrace` + `seccomp`. Notably, where `ptrace` has pre- and post-syscall
events, `ptrace` + `seccomp` only fires a pre-syscall event.
Nonetheless, functionality like switching syscall registers around works
fine in the pre-syscall event.

A smaller distinction as a result of this is that to cancel a syscall, we
need to pass in -1 as the modified syscall, whereas with pure `ptrace` we
would swap the syscall with `getpid` in the post-syscall event. This
distinction is hidden from the Python side (where you can simply pass -1
at all times).

Code that relied both on changing the syscall to perform a different
action *and* changing its return value is now broken when running under
`seccomp`, but I can't think of a reason why anyone would want to do that
-- we certainly dont.

### DIFF
--- a/dmoj/cptbox/helper.cpp
+++ b/dmoj/cptbox/helper.cpp
@@ -134,7 +134,6 @@ int cptbox_child_run(const struct child_config *config) {
 
     kill(getpid(), SIGSTOP);
 
-
 #if PTBOX_SECCOMP
     if (config->trace_syscalls) {
         scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_TRACE(0));
--- a/dmoj/cptbox/helper.cpp
+++ b/dmoj/cptbox/helper.cpp
@@ -26,6 +26,34 @@
 #   define FD_DIR "/proc/self/fd"
 #endif
 
+inline unsigned int get_seccomp_arch(int type) {
+    switch (type) {
+#ifdef SCMP_ARCH_X86
+        case DEBUGGER_X86:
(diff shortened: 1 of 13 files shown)


dmoj/cptbox/helper.cpp
```diff
@@ -103,7 +131,32 @@ int cptbox_child_run(const struct child_config *config) {
 
     if (ptrace_traceme())
         return 204;
+
     kill(getpid(), SIGSTOP);
```
quantum5: Do we want to debug the seccomp initialization with `ptrace`?

Xyene: No trap will be generated until `seccomp` returns when `ptrace` is set up with `PTRACE_O_TRACESECCOMP`. If we move the `SIGSTOP` `kill` until after the `seccomp` initialization, then `cptbox` will receive a syscall trap (for `kill`) before being signaled that everything is "ready" by `SIGSTOP`.

I don't really want to reason if that's OK.

PR #384, opened 2018-12-29
Implement support for FreeBSD 12.0-RELEASE

Xyene:
Addresses #385.


Xyene:
This actually exposes some issues in the FreeBSD syscall list. In particular, `syscalls.py` does not handle the case where there are duplicate entries in the table (for sanity we should immediately error out in these cases) &mdash; we have 3 entries for `fstat`, and the last one overrules all preceding entries.

PR #386, opened 2018-12-31
Implement caching for general executors

Xyene:


dmoj/executors/base_executor.py
```diff
@@ -261,7 +270,53 @@ def get_env(self):
...
+# user's perspective, though, once an Executor is instantiated, it should be ready
+# to launch (e.g. the user shouldn't have to care about compiling themselves). As
+# a compromise, we use a metaclass to compile after all constructors have ran.
+#
+# Using a metaclass also allows us to handle caching executors transparently.
+# Contract: if cached=True is specified and an entry exists in the cache,
+# `create_files` and `compile` will not be run, and `_executable` will be loaded
+# from the cache.
+class CompiledExecutorMeta(type):
+    def _cleanup_cache_entry(key, executor):
```
quantum5: `@staticmethod`?

dmoj/executors/base_executor.py
```diff
@@ -261,7 +270,53 @@ def get_env(self):
...
+                # Minimal sanity checking: is the file still there? If not, we'll just recompile.
+                if os.path.isfile(executor._executable):
+                    obj._executable = executor._executable
+                    return obj
+
+        obj.create_files(*args, **kwargs)
+        obj.compile()
+
+        if is_cached:
+            CompiledExecutorMeta.compiled_binary_cache[cache_key] = obj
```
quantum5: `self.`

dmoj/executors/base_executor.py
```diff
@@ -261,7 +270,53 @@ def get_env(self):
...
+        obj.is_cached = is_cached
+
+        # Before writing sources to disk, check if we have this executor in our cache.
+        if is_cached:
+            cache_key = obj.get_binary_cache_key()
+            if cache_key in CompiledExecutorMeta.compiled_binary_cache:
+                executor = CompiledExecutorMeta.compiled_binary_cache[cache_key]
+                # Minimal sanity checking: is the file still there? If not, we'll just recompile.
+                if os.path.isfile(executor._executable):
+                    obj._executable = executor._executable
```
quantum5: I don't know how I feel about just setting the executable. This will make `self._dir` go out of sync with `self._executable`, as well as not having all the other files created by `create_files`.

Xyene: Alternatives:

* `deepcopy` the `executor` and return it instead (weird because constructor runs once and is then disposed)
* return the `executor` (weird for the same reason)
* copy over `_dir` as well

It's going to be an imperfect solution either way I think, the only question is which weird behaviour we prefer more.

quantum5: I think returning the old executor is what you'd expect from such caching, and would be the least disruptive. However, there are other arguments to the constructor that do not alter the generated executable.

The best solution would probably to copy over `_dir`.

Xyene: Done.

dmoj/executors/gcc_executor.py
```diff
@@ -27,23 +27,27 @@ class GCCExecutor(CompiledExecutor):
...
-        self._fds = fds
-        self._writable = writable
-        self.defines = kwargs.get('defines', [])
+            self.source_paths.append(name)
+
+    def get_binary_cache_key(self):
+        key_components = ([self.problem, self.get_command(), self.get_march_flag()] +
+                          self.get_defines() + self.get_flags() + self.get_ldflags() +
+                          list(self.source_dict.values()))
+        return ''.join(key_components)
```
quantum5: Is it... wise... to have it be so long? Maybe consider SHA384?

quantum5: Also class name? I feel like your current solution could return the wrong class, if say, the flags are controlled by... environment.

dmoj/executors/base_executor.py
```diff
@@ -346,6 +410,7 @@ def get_compile_process(self):
                   'preexec_fn': self.create_executable_limits(), 'time_limit': self.compiler_time_limit}
         kwargs.update(self.get_compile_popen_kwargs())
 
+        log.debug("Compiling: `%s`", ' '.join(self.get_compile_args()))
```
quantum5: Do you actually want this?

dmoj/generator.py
```diff
@@ -6,20 +6,16 @@
...
-
-    def get_generator(self, filename, flags):
+    def get_generator(self, filenames, flags):
         from dmoj.executors import executors
 
-        filename = os.path.abspath(filename)
-        cache_key = filename, tuple(flags)
-        if cache_key in self._cache:
-            return self._cache[cache_key]
+        filenames = map(os.path.abspath, filenames)
```
quantum5: Is this even necessary?

Xyene: I actually had my doubts about that too, but the previous code did it so I replicated it for the multi-file scenario.

dmoj/executors/base_executor.py
Xyene: It's added as a debug-level call, and only the testsuite runs with logging configured for debug. I figured it would be a useful addition.

dmoj/generator.py
```diff
@@ -6,20 +6,16 @@
...
+        sources = {}
 
         try:
-            with open(filename) as file:
-                source = file.read()
+            for filename in filenames:
+                with open(filename, 'r') as f:
```
kiritofeng: From an instance in prod:

Given this `generator` node in the `init.yml`
```
generator:
    source: generator.cpp
    time_limit: 20
    memory_limit: 262144
```

We obtain the following traceback:
```
    File "<redacted>/judge/dmoj/generator.py", line 17, in get_generator
       with open(filename, 'r') as f:
IOError: [Errno 2] No such file or directory: u'<redacted>/problems/peths/ds4/time_limit'
```

dmoj/executors/base_executor.py
xiaowuc1:
```
 File "/code/judge/dmoj/judge.py", line 185, in _begin_grading
   for result in self.grade_cases(grader, problem.cases, short_circuit=short_circuit):
 File "/code/judge/dmoj/judge.py", line 240, in grade_cases
   is_short_circuiting=is_short_circuiting):
 File "/code/judge/dmoj/judge.py", line 264, in grade_cases
   result = grader.grade(case)
 File "/code/judge/dmoj/graders/standard.py", line 33, in grade
 File "/code/judge/dmoj/problem.py", line 225, in input_data
   self._run_generator(gen, args=self.config.generator_args)
 File "/code/judge/dmoj/problem.py", line 185, in _run_generator
   executor = self.problem.generator_manager.get_generator(filenames, flags)
 File "/code/judge/dmoj/generator.py", line 44, in get_generator
   executor = clazz('_generator', None, aux_sources=sources, cached=True)
 File "/code/judge/dmoj/executors/base_executor.py", line 318, in __call__
   self.compiled_binary_cache[cache_key] = obj
 File "/usr/local/lib/python2.7/dist-packages/pylru.py", line 139, in __setitem__
   self.callback(node.key, node.value)
TypeError: 'staticmethod' object is not callable
```

dmoj/problem.py
```diff
@@ -157,25 +157,35 @@ def _run_generator(self, gen, args=None):
...
+            filenames = list(gen.unwrap())
         else:
-            filename = os.path.join(base, gen.source)
+            filenames = gen.source
             if gen.flags:
                 flags += gen.flags
             if not args and gen.args:
                 args += gen.args
 
+
```
quantum5: Why this extra line?

PR #397, opened 2019-03-10
Add support for the Mozilla Spidermonkey Shell.

ehhthing:
Documentation for the JS Shell can be found at https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Introduction_to_the_JavaScript_shell

(I'm not sure if PRs *should* contain new language support, but I suppose it can't hurt.)

### DIFF
--- a/dmoj/executors/MOZJS.py
+++ b/dmoj/executors/MOZJS.py
@@ -1,0 +1,12 @@
+from .base_executor import ScriptExecutor
+
+
+ class Executor(ScriptExecutor):
+    ext = '.js'
+    name = 'MOZJS'
+    command = 'js'
+    test_program = 'print(readline());'
+    nproc = -1
+    @classmethod
+    def get_version_flags(cls, command):
+        return ['-v']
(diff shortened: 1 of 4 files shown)


dmoj/executors/MOZJS.py
```diff
@@ -0,0 +1,12 @@
+from .base_executor import ScriptExecutor
+
+
+ class Executor(ScriptExecutor):
+    ext = '.js'
+    name = 'MOZJS'
+    command = 'js'
+    test_program = 'print(readline());'
+    nproc = -1
+    @classmethod
```
quantum5: New line

quantum5:
I am not sure if I like the SpiderMonkey shell. It's better than the sample v8 shell for sure, but it was vastly underpowered which is why [v8dmoj](/DMOJ/v8dmoj) was created.

This sample shell seems focused on debugging, and only has line based I/O. This might make it hard to do binary input, not that we really have problems that exercise this feature. Still, `v8dmoj` parity or even directly source compatibility would be desirable.

More debate will be necessary before this can be merged.

PR #398, opened 2019-03-15
Optionally allow for checkers to always run

kiritofeng:
Some contests (e.g. ECOO) will check partial output, regardless of if there is a TLE/IR/RTE verdict.

### DIFF
--- a/dmoj/graders/standard.py
+++ b/dmoj/graders/standard.py
@@ -101,18 +101,20 @@ class StandardGrader(BaseGrader):
         # We shouldn't run checkers if the submission is already known to be incorrect, because some checkers
         # might be very computationally expensive.
         # See https://github.com/DMOJ/judge/issues/170
-        if not result.result_flag:
+        checker = case.checker()
+        # checker is a `partial` object, NOT a `function` object
+        if not result.result_flag or getattr(checker.func, 'run_on_error', False):
             # Checkers might crash if any data is None, so force at least empty string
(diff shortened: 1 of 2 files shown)


dmoj/graders/standard.py
```diff
@@ -101,18 +101,19 @@ def check_result(self, case, result):
         # We shouldn't run checkers if the submission is already known to be incorrect, because some checkers
         # might be very computationally expensive.
         # See https://github.com/DMOJ/judge/issues/170
-        if not result.result_flag:
+        checker = case.checker()
+        if not result.result_flag or getattr(checker.func, 'run_on_error', False):
```
Xyene: Did you test this code?

```python
In [1]: def check(foo): 
   ...:     print(foo) 
   ...:                                                                                                                                                                                                                                                                                                                

In [2]: check.run_on_error = True   

In [4]: getattr(check, "run_on_error", False)                                                                                                                                                                                                                                                                          
Out[4]: True

In [9]: getattr(check.func, "run_on_error", False)                                                                                                                                                                                                                                                                     
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-6966974e4b7e> in <module>
----> 1 getattr(check.func, "run_on_error", False)

AttributeError: 'function' object has no attribute 'func'
```

kiritofeng: `check` is a function, but `checker` is a `partial` object

PR #412, opened 2019-04-04
Don't use `utf8bytes` on PY3

kiritofeng:


dmoj/judge.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import os
 import signal
+import six
```
quantum5: `six` is not a standard library module...

dmoj/judge.py
```diff
@@ -423,7 +424,10 @@ def judge_proc(need_monitor):
     logging.basicConfig(filename=logfile, level=logging.INFO,
                         format='%(levelname)s %(asctime)s %(process)d %(module)s %(message)s')
 
-    setproctitle(utf8bytes('DMOJ Judge: %s on %s' % (env['id'], make_host_port(judgeenv))))
+    if six.PY2:
+        setproctitle(utf8bytes('DMOJ Judge: %s on %s' % (env['id'], make_host_port(judgeenv))))
+    else:
+        setproctitle('DMOJ Judge: %s on %s' % (env['id'], make_host_port(judgeenv)))
```
quantum5: Also can you use a variable instead of duplicating code?

PR #416, opened 2019-04-08
Implement data segment grace in cptbox

Xyene:


dmoj/cptbox/sandbox.py
```diff
@@ -156,8 +156,8 @@ def __init__(self, debugger, _, args, executable=None, security=None, time=0, me
         self._cpu_time = time + 5 if time else 0
         self._memory = memory
         self._child_personality = personality
-        self._child_memory = memory * 1024
-        self._child_address = self._child_memory + address_grace * 1024 if memory else 0
+        self._child_memory = memory * 1024 + data_grace * 1024
```
quantum5: I don't like this. This will increase the stack size limit as well.

PR #429, opened 2019-04-16
Improve unittest coverage

Xyene:


dmoj/tests/test_config.py
```diff
@@ -0,0 +1,47 @@
...
+
+        for a, b in enumerate(self.c.list):
+            self.assertEqual(a, b)
+
+    def test_simple_dynamic_keys(self):
+        self.assertEqual(self.c.dynamic1, 10)
+
+    def test_complex_dynamic_keys(self):
+        self.assertEqual(self.c.dynamic2.foo, 1)
+        self.assertEqual(self.c.dynamic2.bar.unwrap(), [0, 1])
```
quantum5: On an unrelated note... perhaps we should define `__eq__` on `ConfigNode`.

PR #431, opened 2019-04-18
Fix `process.returncode` comparison

kiritofeng:


dmoj/graders/standard.py
```diff
@@ -138,12 +138,9 @@ def set_result_flag(self, process, result):
                 result.result_flag |= Result.RTE
                 process.signal = signal.SIGSEGV
             else:
-                # print>> sys.stderr, 'Exited with error: %d' % process.returncode
                 result.result_flag |= Result.IR
+        # None < 0 == True
```
quantum5: Dude, `None < 0` is invalid on Python 3, the only version we care about now.

quantum5:
I think the PR should be renamed "Fix None and int comparison" or something, since it actually fixes a bug!

PR #444, opened 2019-05-06
Make Go Executor PY3 compliant

kiritofeng:


Xyene:
Can we get a test testing the package/comment stripping behaviour?

PR #445, opened 2019-05-07
cptbox: Unconditionally disallow writing to files

kiritofeng:
This breaks IO symlinks; this will be addressed in a future commit.


dmoj/cptbox/chroot.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import print_function
```
kiritofeng: There are no `print` calls in this file.

PR #447, opened 2019-05-15
GO: Strip block comments before validating package

Xyene:


dmoj/executors/GO.py
```diff
@@ -4,8 +4,9 @@
 from dmoj.error import CompileError
 from .base_executor import CompiledExecutor
 
-recomment = re.compile(br'//.*?(?=[\r\n])')
-decomment = lambda x: recomment.sub(b'', x)
+reinline_comment = re.compile(r'//.*?(?=[\r\n])', re.U)
+recomment = re.compile(r'/\*.*?\*/', re.DOTALL | re.U)
+decomment = lambda x: reinline_comment.sub(b'', recomment.sub(b'', x))
```
kiritofeng: You need to make `x` a `string` object, and the empty string shouldn't be bytestrings.

PR #461, opened 2019-07-24
Show generator stderr in case of non-zero return code

### LINKED ISSUE #362: Include generator feedback in InternalError
Instead of just including the raw error code, we should also include the feedback if it exists (`proc.feedback if hasattr(proc, 'feedback')`). This would be more optimal than memorizing common signal codes.

https://github.com/DMOJ/judge/blob/77dd0984781b56ad870d7541f6f591e2c564fa8a/dmoj/problem.py#L229

kiritofeng:
Closes #362.

### DIFF
--- a/dmoj/result.py
+++ b/dmoj/result.py
@@ -58,6 +58,16 @@ class Result(object):
     def output(self):
         return utf8text(self.proc_output[:self.case.output_prefix_length], 'replace')
 
+    def set_result_flag(self, process):
+        if process.returncode > 0:
+            self.result_flag |= Result.IR
+        if process.returncode is None or process.returncode < 0:
+            self.result_flag |= Result.RTE  # Killed by signal
+        if process.tle:
+            self.result_flag |= Result.TLE
+        if process.mle:
+            self.result_flag |= Result.MLE
+
 
 class CheckerResult(object):
     def __init__(self, passed, points, feedback=None, extended_feedback=None):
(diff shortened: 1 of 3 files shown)


dmoj/result.py
```diff
@@ -1,8 +1,29 @@
+import os
+import signal
 import six
 
 from dmoj.utils.unicode import utf8text
 
 
+def set_result_flag(process, result):
```
quantum5: Why not make this a method on `Result`? It's clearly associated with that class...

Xyene:
From CI,

```
Start grading generator_test_mle/22 in CPP11...
  File "/home/travis/build/DMOJ/judge/dmoj/judge.py", line 187, in _begin_grading
    for result in self.grade_cases(grader, grader.cases(), short_circuit=short_circuit):
  File "/home/travis/build/DMOJ/judge/dmoj/judge.py", line 266, in grade_cases
    result = grader.grade(case)
  File "/home/travis/build/DMOJ/judge/dmoj/graders/standard.py", line 34, in grade
  File "/home/travis/build/DMOJ/judge/dmoj/problem.py", line 231, in input_data
    self._run_generator(gen, args=self.config.generator_args)
  File "/home/travis/build/DMOJ/judge/dmoj/problem.py", line 222, in _run_generator
    raise InternalError(error)
InternalError: generator exited with nonzero code -6 with feedback: 
```

I would have expected this to raise with something like `InternalError: generator exited with nonzero code -6 with feedback: std::bad_alloc`, or at least not an empty string. Is this correct?

PR #466, opened 2019-08-31
Pass `stdin` and `stdout` arguments to SecurePopen if available

Ninjaclasher:


dmoj/executors/mixins.py
```diff
@@ -156,7 +156,9 @@ def launch(self, *args, **kwargs):
                                    time=kwargs.get('time'),
                                    memory=kwargs.get('memory'),
                                    wall_time=kwargs.get('wall_time'),
-                                   stderr=(PIPE if kwargs.get('pipe_stderr', False) else None),
+                                   stdin=kwargs.get('stdin', PIPE),
```
Xyene: The defaults for all of these in `subprocess` are `None`, so we probably want to be symmetric with that: https://docs.python.org/3/library/subprocess.html#subprocess.run

PR #479, opened 2019-09-16
Remove Windows support from the judge

Xyene:


dmoj/executors/mixins.py
```diff
@@ -10,158 +10,117 @@
...
+                               time=kwargs.get('time'),
+                               memory=kwargs.get('memory'),
+                               wall_time=kwargs.get('wall_time'),
+                               stdin=kwargs.get('stdin'),
+                               stdout=kwargs.get('stdout'),
+                               stderr=kwargs.get('stderr'),
+                               env=env, cwd=utf8bytes(self._dir),
+                               nproc=self.get_nproc(),
+                               fsize=self.fsize)
 except ImportError:
```
quantum5: Why are we still doing this? Let's just have the *nix `PlatformExecutorMixin` directly in the global scope. Then if things fail, you'd know what `cptbox` error it is instead of `NameError` on `PlatformExecutorMixin`.

dmoj/generator.py
```diff
@@ -42,13 +42,6 @@ def find_runtime(languages):
             compiler_time_limit = compiler_time_limit or clazz.compiler_time_limit
             clazz = type('Executor', (clazz,), {'compiler_time_limit': compiler_time_limit})
 
-        if hasattr(clazz, 'flags'):
-            flags += ['-DWINDOWS_JUDGE', '-DWIN32'] if os.name == 'nt' else ['-DLINUX_JUDGE']
```
quantum5: Shouldn't we still `-DLINUX_JUDGE` for compat?

setup.py
```diff
@@ -5,14 +5,13 @@
 import os
 import sys
 
-if sys.version_info[0] >= 3 and os.name == 'nt':
-    print('DMOJ is unsupported on Windows Python 3, please use Python 2 instead.', file=sys.stderr)
+if os.name == 'nt':
+    print('DMOJ is unsupported on Windows.', file=sys.stderr)
     sys.exit(0)
```
quantum5: I think perhaps this should fail with an error exit...

dmoj/generator.py
Xyene: We have precisely 2 problems (`linked` and `ds4`) that rely on it. I don't think it's worth maintaining this hack just for them.

kiritofeng: `ds4` has it to maintain parity with the judge's compiler so people don't attempt something like
```cpp
#ifdef LINUX_JUDGE
#include<bits/extc++.h>
#endif
```

quantum5: What? Why would the generator matter at all?

Issue #478, opened 2019-09-15
Move ANSI HTML-ifying to the site


Xyene:
The judge [formats compiler ANSI codes as HTML](https://github.com/DMOJ/judge/blob/master/dmoj/utils/ansi.py#L16), which are then sent to the site, where they rendered as "safe" (unescaped).

Some problems with this were recognized from the start, but the ease of implementation all those years ago outweighed doing it right. In particular,

* Judges have to be trusted to not output things like `script` tags into the compiler output section
* There is no future control over the mapping from ANSI code to color; this makes wishlist items like a dark mode significantly harder to implement
* There is some (small) database bloat from storing HTML tags rather than ANSI codes

Moving forward, we should fix this. A possible solution is to write a migration that strips HTML tags from old compiler outputs, and start sending ANSI codes instead.

Xyene:
This is complete and live.

PR #480, opened 2019-09-16
Move ANSI HTML-ifying to the site

Xyene:
Addresses #478.

Should not be merged until the appropriate site-side change is made.

### DIFF
--- a/dmoj/judge.py
+++ b/dmoj/judge.py
@@ -173,7 +173,8 @@ class Judge(object):
         self.current_submission_thread.daemon = True
         self.current_submission_thread.start()
 
-        if blocking:
+        # Submission may have been killed already, but block if it hasn't been.
+        if blocking and self.current_submission_thread is not None:
             self.current_submission_thread.join()
 
     def grade_cases(self, grader, cases, short_circuit=False, is_short_circuiting=False):
--- a/dmoj/judge.py
+++ b/dmoj/judge.py
@@ -635,7 +635,7 @@ def main():  # pragma: no cover
     print('Running live judge...')
 
     for warning in judgeenv.startup_warnings:
-        print(ansi_style('#ansi[Warning: %s](yellow)' % warning))
+        print_ansi('#ansi[Warning: %s](yellow)' % warning)
     del judgeenv.startup_warnings
 
     if 'judges' in env:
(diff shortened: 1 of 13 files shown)


dmoj/utils/ansi.py
```diff
@@ -29,3 +26,7 @@ def format_inline(text, attrs):
 
     return re.sub(r'#ansi\[(.*?)\]\((.*?)\)',
                   lambda x: format_inline(x.group(1), x.group(2)) if not no_ansi else x.group(1), text)
+
+
+def print_ansi(*args, **kwargs):
+    print(*list(map(ansi_style, args)), **kwargs)
```
quantum5: Do you need the `list`?

```pycon
$ python3
Python 3.7.3 (default, Apr  3 2019, 05:39:12) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> print(*map(int, '123'))
1 2 3
```

Xyene:
> 
> 
> I don't see how `print_ansi` is actually relevant to this change. Perhaps it would be better to make it a separate PR...

Perhaps, but it was bugging my autoformatter and this'll be rebased anyway.

dmoj/utils/ansi.py
```diff
@@ -29,3 +26,7 @@ def format_inline(text, attrs):
 
     return re.sub(r'#ansi\[(.*?)\]\((.*?)\)',
                   lambda x: format_inline(x.group(1), x.group(2)) if not no_ansi else x.group(1), text)
+
+
+def print_ansi(*args, **kwargs):
+    print(*map(ansi_style, args), **kwargs)
```
quantum5: Wait, what happens if there are non-string arguments?

dmoj/judge.py
```diff
@@ -183,7 +183,8 @@ def begin_grading(self, id, *args, **kwargs):
         self.current_submission_thread.daemon = True
         self.current_submission_thread.start()
 
-        if is_blocking:
+        # Submission may have been killed already, but block if it hasn't been.
```
quantum5: What does this have to do with ANSI escape codes?

Xyene: It's a race. Not HTMLifying makes it trigger all the time and crashes the judge, so it is fixed here.

dmoj/utils/ansi.py
Xyene: There are currently none, and I don't want to introduce more unused code. It's trivially fixable if we ever need to print non-string things.

PR #484, opened 2019-09-17
Pass `-fmax-errors=N` to GCC/Clang; #471

### LINKED ISSUE #471: Pass `-fmax-errors=N` to GCC/Clang
For some reasonable choice of `N` (3? 5?), this should save us some excessive database bloat with really wrong submissions.

It's also a bit more friendly for users, who will at least see some part of their error message, rather than just an unhelpful "compiler output too long" message.

kiritofeng:
Closes #471.

### DIFF
--- a/dmoj/executors/gcc_executor.py
+++ b/dmoj/executors/gcc_executor.py
@@ -9,6 +9,7 @@ from .base_executor import CompiledExecutor
 
 GCC_ENV = env.runtime.gcc_env or {}
 GCC_COMPILE = os.environ.copy()
+MAX_ERRORS = 5
 
 if os.name == 'nt':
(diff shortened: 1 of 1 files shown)


dmoj/executors/gcc_executor.py
```diff
@@ -22,7 +23,7 @@
 
 class GCCExecutor(CompiledExecutor):
     defines = []
-    flags = []
+    flags = ['-fmax-errors=%d' % MAX_ERRORS]
```
quantum5: This feels like an awful idea. The whole point of `flags` here is so that subclasses can override it. We don't actually have a metaclass to inherit by concatenation, nor can we expect subclasses to write `flags = GCCExecutor.flags + [...]`.

PR #486, opened 2019-09-19
Fix #485

### LINKED ISSUE #485: java_executor throws an exception when judging an interactive problem where the program crashes
Production reproduction: https://dmoj.ca/submission/1595554

```
 File "/code/judge/dmoj/graders/custom.py", line 15, in grade
   return self._grader.grade(case)
 File "/code/judge/dmoj/graders/standard.py", line 66, in grade
   self.update_feedback(check, error, process, result)
 File "/code/judge/dmoj/graders/standard.py", line 83, in update_feedback
   getattr(self.binary, 'get_feedback', lambda x, y, z: '')(error, result, process)))
 File "/code/judge/dmoj/executors/java_executor.py", line 121, in get_feedback
   if b'Error: Main method not found in class' in stderr:
TypeError: argument of type 'NoneType' is not iterable
```

Xyene:
Fixes #485.

### DIFF
--- a/dmoj/cptbox/sandbox.py
+++ b/dmoj/cptbox/sandbox.py
@@ -328,7 +328,7 @@ class SecurePopen(six.with_metaclass(SecurePopenMeta, Process)):
 
         if stdin == PIPE:
             self._child_stdin, self._stdin = os.pipe()
-            self.stdin = os.fdopen(self._stdin, 'w')
+            self.stdin = os.fdopen(self._stdin, 'wb')
         elif isinstance(stdin, int):
             self._child_stdin, self._stdin = stdin, -1
         elif stdin is not None:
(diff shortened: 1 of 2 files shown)


dmoj/cptbox/sandbox.py
```diff
@@ -328,7 +328,7 @@ def __init_streams(self, stdin, stdout, stderr):
 
         if stdin == PIPE:
             self._child_stdin, self._stdin = os.pipe()
-            self.stdin = os.fdopen(self._stdin, 'w')
+            self.stdin = os.fdopen(self._stdin, 'wb')
```
kiritofeng: Passing the binary flag causes the writers to change from `_io.TextIOWrapper` to `_io.BufferedWriter`, which does not actually write the data...

quantum5: Huh? Everyone knows that if you don't want buffering, pass `buffering=0`.

PR #503, opened 2019-09-30
Use `Decimal` in floats checker

### LINKED ISSUE #456: checker improvement: make floats checker use arbitrary-precision Decimal
Most floating-point checkers do not depend on this, however it can avoid issues when the precision guarantee needed is very tight.

kiritofeng:
Closes #456.

### DIFF
--- a/dmoj/checkers/floats.py
+++ b/dmoj/checkers/floats.py
@@ -1,17 +1,21 @@
+from decimal import Decimal
 from re import split as resplit
+from typing import Union
 
 from dmoj.error import InternalError
 from dmoj.utils.unicode import utf8bytes
 
 
-def verify_absolute(process_float: float, judge_float: float, epsilon: float) -> bool:
+def verify_absolute(process_float: Union[float, Decimal], judge_float: Union[float, Decimal],
+                    epsilon: Union[float, Decimal]) -> bool:
     # Since process_float can be NaN, this is NOT equivalent to
     # (process_float - judge_float) > epsilon;  the code below will always
     # reject NaN, even if judge_float is NaN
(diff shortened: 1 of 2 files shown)


dmoj/checkers/floats.py
```diff
@@ -27,7 +28,13 @@ def verify_default(process_float, judge_float, epsilon):
             abs(1.0 - process_float / judge_float) <= epsilon)
 
 
-def check(process_output, judge_output, precision=6, error_mode='default', **kwargs):
+def check(process_output, judge_output, precision=6, fast=True, error_mode='default', **kwargs):
+
```
Xyene: Remove newline

dmoj/checkers/floats.py
```diff
@@ -27,7 +28,13 @@ def verify_default(process_float, judge_float, epsilon):
...
+    if fast:
```
Xyene: Having people write `fast: False` is kind of a weird way to say that they want arbitrary-precision calculations...

dmoj/checkers/floats.py
```diff
@@ -27,7 +28,12 @@ def verify_default(process_float, judge_float, epsilon):
             abs(1.0 - process_float / judge_float) <= epsilon)
 
 
-def check(process_output, judge_output, precision=6, error_mode='default', **kwargs):
+def check(process_output, judge_output, precision=6, high_precision=False, error_mode='default', **kwargs):
```
quantum5: What about precision and other stuff you can control in the [decimal context](https://docs.python.org/3/library/decimal.html#context-objects)?

PR #509, opened 2019-10-03
make PlatformExecutorMixin an ABC

xiaowuc1:


dmoj/executors/base_executor.py
```diff
@@ -18,9 +18,7 @@
 
 
 class BaseExecutor(PlatformExecutorMixin):
-    ext = None
     nproc = 0
-    command = None
     command_paths = []
     runtime_dict = env.runtime
     name = '(unknown)'
```
Xyene: This definitely exists as-is because of no ABC; it's assigned everywhere.

dmoj/executors/mixins.py
```diff
@@ -35,7 +36,7 @@
     BASE_FILESYSTEM += [r'/etc/ld\.so\.(?:nohwcap|preload|cache)$']
 
 
-class PlatformExecutorMixin(object):
+class PlatformExecutorMixin(abc.ABCMeta):
```
Xyene: Python 3 syntax is `metaclass=...`

dmoj/executors/compiled_executor.py
```diff
@@ -98,7 +102,7 @@ def _shocker_thread(self):
             time.sleep(0.25)
 
 
-class CompiledExecutor(BaseExecutor, metaclass=_CompiledExecutorMeta):
+class CompiledExecutor(CompiledExecutorMeta):
```
quantum5: We are no longer using the metaclass.

dmoj/executors/compiled_executor.py
```diff
@@ -63,6 +63,10 @@ def __call__(self, *args, **kwargs):
         return obj
 
 
+class CompiledExecutorMeta(BaseExecutor, _CompiledExecutorMeta):
```
quantum5: This is a normal class mixed in with a metaclass.

dmoj/executors/compiled_executor.py
```diff
@@ -24,7 +25,7 @@
 # Contract: if cached=True is specified and an entry exists in the cache,
 # `create_files` and `compile` will not be run, and `_executable` will be loaded
 # from the cache.
-class _CompiledExecutorMeta(type):
+class _CompiledExecutorMeta(type, metaclass=abc.ABCMeta):
```
quantum5: I don't think you need `metaclass=abc.ABCMeta` here.

dmoj/executors/compiled_executor.py
```diff
@@ -63,6 +64,14 @@ def __call__(self, *args, **kwargs):
         return obj
 
 
+# Because BaseExecutor has ABCMeta as a metaclass, we create this class so
+# CompiledExecutor can have a proper metaclass. See
+# http://www.phyast.pitt.edu/~micheles/python/metatype.html for more
+# details.
+class _BaseExecutorCompiledExecutorMeta(abc.ABCMeta, _CompiledExecutorMeta):
```
quantum5: I think you don't need class. Just make `_CompiledExecutorMeta` inherit from `abc.ABCMeta`, which has to implicitly inherit from `type`. We don't have any other users of `_CompiledExecutorMeta`.

PR #515, opened 2019-10-03
Generators should not be subject to OLE

Xyene:


dmoj/problem.py
```diff
@@ -193,13 +193,14 @@ def _run_generator(self, gen, args=None):
         else:
             proc = executor.launch_unsafe(*args, stdin=subprocess.PIPE, stdout=subprocess.PIPE,
                                           stderr=subprocess.PIPE)
+            proc.unsafe_communicate = proc.communicate
```
Xyene: This is a hack I don't particularly like, maybe we should just commit to having all generators sandboxed and call it a day. Long gone are the days where their performance would be too negatively impacted by sandboxing.

PR #517, opened 2019-10-03
Add mypy GitHub action

Xyene:


.github/workflows/mypy.yml
```diff
@@ -0,0 +1,19 @@
...
+      with:
+        python-version: 3.7
+    - name: Install dependencies and mypy
+      run: |
+        pip install cython mypy
+        pip install -r requirements.txt
+    - name: Verify mypy types
+      run: |
+        mypy --version
+        mypy --ignore-missing-imports .
```
xiaowuc1: I think at some point we'll have a `mypy.ini` file that we can use to customize what options we run with - that is possibly more sustainable?

Xyene: Sounds good to me.

PR #520, opened 2019-10-04
changed to PollingObserver for network mount

kailashnath1998:
While using network mounts as problem root the filesystem events don't always get emitted, to detect changes in such cases using ```watchdog.observers.polling.PollingObserver``` instead of the Observer.

Signed-off-by: Kailashnath Nagendran <kailashnath1998@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,8 @@ optional arguments:
                         file to load judge configurations from
   -l LOG_FILE, --log-file LOG_FILE
                         log file to use
-  --no-watchdog         disable use of watchdog on problem directories
+  --no-watchdog         disable use of watchdog on problem directories for nfs
+  --polling-observer    use polling observer instead of observer
   -a API_PORT, --api-port API_PORT
                         port to listen for the judge API (do not expose to
                         public, security is left as an exercise for the
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ optional arguments:
   -l LOG_FILE, --log-file LOG_FILE
                         log file to use
   --no-watchdog         disable use of watchdog on problem directories for nfs
-  --polling-observer    use polling observer instead of observer
+  --use-polling-observer    use polling instead of inotify for problem updates
   -a API_PORT, --api-port API_PORT
                         port to listen for the judge API (do not expose to
                         public, security is left as an exercise for the
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ optional arguments:
                         file to load judge configurations from
   -l LOG_FILE, --log-file LOG_FILE
                         log file to use
-  --no-watchdog         disable use of watchdog on problem directories for nfs
+  --no-watchdog         disable use of watchdog on problem directories
   --use-polling-observer    use polling instead of inotify for problem updates
   -a API_PORT, --api-port API_PORT
                         port to listen for the judge API (do not expose to
--- a/dmoj/judgeenv.py
+++ b/dmoj/judgeenv.py
@@ -41,7 +41,7 @@ env = ConfigNode(defaults={
 }, dynamic=False)
 _root = os.path.dirname(__file__)
 
-log_file = server_host = server_port = no_ansi = no_watchdog = problem_regex = case_regex = None
+log_file = server_host = server_port = no_ansi = no_watchdog = polling_observer = problem_regex = case_regex = None
 secure = no_cert_check = False
 cert_store = api_listen = None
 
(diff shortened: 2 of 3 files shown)


Xyene:
Thanks for PRing!

While I think this change might be useful for some setups involving small problemsets over network mounts, it shouldn't be the default. On [dmoj.ca](https://dmoj.ca) we mount data over NFS, and polling north of 2000 problems is prohibitively expensive. The solution in cases like these is to run a daemon on the NFS _server_ that calls the [judge API endpoint](https://github.com/DMOJ/judge/blob/master/dmoj/control.py#L4) to force it to refresh its list. If one judge also runs the NFS server for other judges, it just needs to declare [`update_pings`](https://github.com/DMOJ/judge/blob/bfcdbe5e1fcbd7ac32e021c92cb9d161a97e01ca/dmoj/judgeenv.py#L37) in its YAML.

So, I think this change is good in reducing setup overhead in some use-cases involving small problemsets, but should be behind a command-line flag.

kailashnath1998:
@Xyene thanks for pointing out the issues, have made necessary changes to add polling observer as command-line flag

dmoj/judgeenv.py
```diff
@@ -86,6 +86,8 @@ def load_env(cli=False, testsuite=False):  # pragma: no cover
                                  'security is left as an exercise for the reverse proxy)')
         parser.add_argument('-A', '--api-host', default='127.0.0.1',
                             help='IPv4 address to listen for judge API')
+        parser.add_argument('--polling-observer', action='store_true',
```
Xyene: `--use-polling-observer`

dmoj/monitor.py
```diff
@@ -70,7 +71,11 @@ def __init__(self):
                 self._refresher = None
 
             self._handler = SendProblemsHandler(self._refresher)
-            self._monitor = monitor = Observer()
+            if judgeenv.polling_observer:
+                self._monitor = monitor = PollingObserver()
+            else:
+                self._monitor = monitor = Observer()
```
Xyene: For style, change this to assign only `self._monitor` and change the only usage of `monitor` below to be `self._monitor` as well.

dmoj/judgeenv.py
```diff
@@ -86,6 +86,8 @@ def load_env(cli=False, testsuite=False):  # pragma: no cover
...
+                            help='use polling observer instead of observer')
```
Xyene: Maybe be more explicit, e.g. "use polling instead of inotify for problem updates", so that the reader doesn't need to know about the implementation details (`watchdog`'s `Observer` class) to use this flag.

kailashnath1998:
All suggested changes have been made

dmoj/monitor.py
```diff
@@ -70,7 +71,12 @@ def __init__(self):
                 self._refresher = None
 
             self._handler = SendProblemsHandler(self._refresher)
-            self._monitor = monitor = Observer()
+            if judgeenv.polling_observer:
+                self._monitor = PollingObserver()
+            else:
+                self._monitor = Observer()
+
+            monitor = self._monitor
```
Xyene: I meant inlining this into the `for` loop before.

README.md
```diff
@@ -140,6 +140,7 @@ optional arguments:
   -l LOG_FILE, --log-file LOG_FILE
                         log file to use
   --no-watchdog         disable use of watchdog on problem directories
+  --use-polling-observer    use polling instead of inotify for problem updates
```
quantum5: Maybe put the help text on a new line like what we do for `-a API_PORT, --api-port API_PORT`.

Xyene: That is, `dmoj --help` with this patch writes:

```
$ dmoj --help
usage: dmoj [-h] [-p SERVER_PORT] [-c CONFIG] [-l LOG_FILE] [--no-watchdog]
            [-a API_PORT] [-A API_HOST] [--use-polling-observer] [-s] [-k]
            [-T TRUSTED_CERTIFICATES]
            [-e ONLY_EXECUTORS | -x EXCLUDE_EXECUTORS] [--no-ansi]
            server_host [judge_name] [judge_key]

Spawns a judge for a submission server.

positional arguments:
  server_host           host to connect for the server
  judge_name            judge name (overrides configuration)
  judge_key             judge key (overrides configuration)

optional arguments:
  -h, --help            show this help message and exit
  -p SERVER_PORT, --server-port SERVER_PORT
                        port to connect for the server
  -c CONFIG, --config CONFIG
                        file to load judge configurations from (default:
                        ~/.dmojrc)
  -l LOG_FILE, --log-file LOG_FILE
                        log file to use
  --no-watchdog         disable use of watchdog on problem directories
  -a API_PORT, --api-port API_PORT
                        port to listen for the judge API (do not expose to
                        public, security is left as an exercise for the
                        reverse proxy)
  -A API_HOST, --api-host API_HOST
                        IPv4 address to listen for judge API
  --use-polling-observer
                        use polling instead of inotify for problem updates
  -s, --secure          connect to server via TLS
  -k, --no-certificate-check
                        do not check TLS certificate
  -T TRUSTED_CERTIFICATES, --trusted-certificates TRUSTED_CERTIFICATES
                        use trusted certificate file instead of system
  -e ONLY_EXECUTORS, --only-executors ONLY_EXECUTORS
                        only listed executors will be loaded (comma-separated)
  -x EXCLUDE_EXECUTORS, --exclude-executors EXCLUDE_EXECUTORS
                        prevent listed executors from loading (comma-
                        separated)
  --no-ansi             disable ANSI output
```

PR #525, opened 2019-10-05
Limit Clang/++ compile errors; #487

### LINKED ISSUE #487: Clang and Clang++ always give a compilation warning due to -fmax-errors
The warning in particular is:

```
clang: warning: argument unused during compilation: '-fmax-errors=5' [-Wunused-command-line-argument]
```

Clang's version of this flag is `ferror-limit`.

Carson-Tang:
Closes #487

### DIFF
--- a/dmoj/executors/clang_executor.py
+++ b/dmoj/executors/clang_executor.py
@@ -1,8 +1,11 @@
-from .gcc_executor import GCCExecutor
+from .gcc_executor import GCCExecutor, MAX_ERRORS
 
 
 class ClangExecutor(GCCExecutor):
     arch = 'clang_target_arch'
+
+    def get_flags(self):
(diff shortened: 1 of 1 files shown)


dmoj/executors/clang_executor.py
```diff
@@ -1,9 +1,15 @@
 from .gcc_executor import GCCExecutor
 
 
+MAX_ERRORS = 5
```
Xyene: This should be imported from `.gcc_executor`.

dmoj/executors/clang_executor.py
```diff
@@ -1,9 +1,13 @@
 from .gcc_executor import GCCExecutor
+from .gcc_executor import MAX_ERRORS
```
Xyene: This should be imported as part of the first line.

PR #538, opened 2019-10-09
TUR: give tprologc a fake version

Xyene:
This is so that site-side it may properly appear in the version matrix.


Xyene:
Superseded by https://github.com/DMOJ/site/issues/1113.

PR #543, opened 2019-10-10
Emulate xterm everywhere

Xyene:


dmoj/executors/compiled_executor.py
```diff
@@ -147,11 +148,50 @@ def limit_executable():
...
-                  'preexec_fn': self.create_executable_limits(), 'time_limit': self.compiler_time_limit}
-        kwargs.update(self.get_compile_popen_kwargs())
+        # Some languages may insist on providing certain functionality (e.g. colored highlighting of errors) if they
+        # feel they are connected to a terminal. Some are more persistent than others in enforcing this, so this hack
+        # aims to provide a convincing-enough lie to the runtime so that it starts singing in color.
+        #
+        # Emulate the streams of a process connected to a terminal: stdin, stdout, and stderr are all ptys.
+        self._master, self._slave = pty.openpty()
+        # Some runtimes *cough cough* Swift *cough cough* actually check the environment variables too.
+        env = self.get_compile_env() or os.environ.copy()
```
quantum5: This is actually broken, although you are just copying code. The intention was that `get_compile_env` could return `{}`, which is supposed to mean blank environment, except it actually has the same meaning as `None` now. This is probably not worth changing now.

Xyene: Yeah, I noticed this too.

PR #557, opened 2019-10-13
Prints runtime version during self-testing; #545

### LINKED ISSUE #545: Print runtime version during self-testing
This will ensure that we never break version reporting, and bump up our coverage a bit.

Carson-Tang:
Closes #545


Carson-Tang:
Preview with success and failed builds. Note the extra spacing after "Failed" to align with the feedback from the "Success" builds.
image

Carson-Tang:
Updated look by request of @Xyene 
image

PR #560, opened 2019-10-14
Drop outdated Nim, Octave, Clozure CL, Ruby 1.8/1.9, Java 7, and C++0x executors

Xyene:


Xyene:
That ought to be done in a separate PR, since it probably requires a migration.

PR #562, opened 2019-10-19
Make Travis build on Python 3.8

quantum5:


.travis.yml
```diff
@@ -7,7 +7,7 @@ env:
   matrix:
   - PYTHON_VERSION=3.6 DMOJ_USE_SECCOMP="yes"
   - PYTHON_VERSION=3.7 DMOJ_USE_SECCOMP="yes"
-  - PYTHON_VERSION=3.6 DMOJ_USE_SECCOMP="no"
+  - PYTHON_VERSION=3.8 DMOJ_USE_SECCOMP="yes"
```
Xyene: The fact that this is out of order hurts

quantum5: I feel like seccomp grouping > version grouping. Do you want it to be re-ordered?

PR #565, opened 2019-10-24
Implement testcase format specifiers; closes #262

### LINKED ISSUE #262: Testcase format specifiers
We need an efficient way to specify testcases in a common format for a contest, where dynamic keys are not an optimal solution. For example, the regex `\w+\.(?<batch>[0-9]+)-(?<case>[0-9]+)\.in` matches most, if not all, COCI data.

That's a lot of data.

That's a lot of lines of `init.yml` saved.

There are probably more broad ways to approach this, but for the general case a simple regex would save a lot of work, even if it loses flexibility for more complex cases (dynamic keys are there for those!)

The following should be valid, after this ticket:

```yml
test_format_in: \w+\.(?<batch>[0-9]+)-(?<case>[0-9]+)\.in
test_format_out: \w+\.(?<batch>[0-9]+)-(?<case>[0-9]+)\.out
point_values: [10, 10, 20, 60]
```

Someday this may be generalized to support setting a format specifier for a specific `test_cases` node, but that is beyond the scope of this ticket. For this ticket, if `test_format_(in|out)` exists, then the `test_cases` node should be ignored.

Xyene:


dmoj/problem.py
```diff
@@ -45,6 +51,88 @@ def __init__(self, problem_id, time_limit, memory_limit):
...
+                return None
+
+            # Allow batches and case numbers to be alphanumeric, in which case we will sort them lexicographically.
+            # Still attempt to process them as integers first, though, since most problems will use this format.
+            return try_match_int(match, 'batch'), try_match_int(match, 'case')
+
+        # Match all cases with the same (batch, position) mapping.
+        groups = defaultdict(list)
+        batch_ids = set()
+        for a in filenames:
```
quantum5: Can we not do O(n²)? This should not be rocket science for a grade 10 student. This code has to run for every submission...

dmoj/problem.py
```diff
@@ -45,6 +51,90 @@ def __init__(self, problem_id, time_limit, memory_limit):
...
+            if batch_or_case_id in batch_ids:
+                test_cases.append({
+                    'batched': [{
+                        'in': input_file,
+                        'out': output_file,
+                    } for _, input_file, output_file in sorted(groups[batch_or_case_id])],
+                    'points': next(case_points),
+                })
+            else:
+                assert len(groups[batch_or_case_id]) == 1
```
quantum5: Maybe we should raise real exceptions. Imagine running the judge with `python -O`.

dmoj/problem.py
```diff
@@ -45,6 +51,88 @@ def __init__(self, problem_id, time_limit, memory_limit):
...
+    def _match_test_cases(self, filenames, input_case_pattern, output_case_pattern, case_points):
+        def try_match_int(match, group):
+            try:
+                val = match.group(group)
+            except IndexError:
+                return None
+
+            try:
+                return int(val)
+            except (ValueError, TypeError):
```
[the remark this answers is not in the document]
Xyene: TypeError is raised if `val` is None.

Issue #408, opened 2019-03-25
Buffer testcase updates in network protocol


Xyene:
Currently, when a solution gets WA on a problem with 1000 testcases in a batch, 1000 packets are sent at once to the bridge. The bridge then hammers MySQL with 1000 queries all at once. Other times, submissions grade extremely quickly and achieve the same effect. This has led to load problems in the past, and is likely to lead to load problems in the future -- the bridge is the bottleneck to grading, after all.

The judge should buffer testcase updates that occurred in the last _N_ (_N_ < 1) seconds, and send them to the bridge in a single packet. The bridge could then `bulk_create` the `SubmissionTestCase` models, ideally reducing load.

PR #568, opened 2019-10-26
Implement batching of testcase updates; #408

Xyene:


dmoj/packet.py
```diff
@@ -162,15 +166,51 @@ def _read_single(self) -> dict:
...
-    def run_async(self):
-        threading.Thread(target=self._read_async).start()
+    def _flush_testcase_queue(self):
+        with self._testcase_queue_lock:
+            if not self._testcase_queue:
+                return
+
+            # Copy and release the lock quickly so that submission can continue grading.
+            testcase_queue = self._testcase_queue[:]
+            self._testcase_queue.clear()
```
quantum5: Why this instead of `testcase_queue = self._testcase_queue; self._testcase_queue = []`?

dmoj/packet.py
```diff
@@ -162,15 +166,51 @@ def _read_single(self) -> dict:
...
+
+        self._send_packet({'name': 'test-case-status',
```
quantum5: Race condition: this has to be in the lock. Imagine the background thread flushing, and gets preempted here. Then, the main grading thread ends a batch, calling this function, finding nothing in the queue, sends the batch end, then this thread resumes.

PR #573, opened 2019-11-01
Test ARM64 judge on Travis in Docker

Xyene:


.docker.test.py
```diff
@@ -1,10 +1,21 @@
...
+arch = os.getenv('ARCH', 'amd64')
+if arch == 'amd64':
+    ALLOW_FAIL = {'GASARM', 'JAVA9', 'JAVA10', 'OBJC'}
+    EXECUTORS = get_available()
+elif arch == 'aarch64':
+    EXECUTORS = {'AWK', 'BF', 'C', 'C11', 'CPP03', 'CPP11', 'CPP14', 'CPP17',
+                 'JAVA8', 'JAVA11', 'PAS', 'PERL', 'PY2', 'PY3', 'SED', 'TEXT'}
+    ALLOW_FAIL = set()
+else:
+    raise
```
quantum5: Huh? What's it raising?

.docker.test.py
```diff
@@ -1,10 +1,21 @@
 import logging
+import os
 
 from dmoj import judgeenv
 from dmoj.citest import ci_test
 from dmoj.executors import get_available
 
-ALLOW_FAIL = {'GASARM', 'JAVA9', 'JAVA10', 'OBJC'}
+arch = os.getenv('ARCH', 'amd64')
```
quantum5: Why not use `platform.machine()`? You can't screw that up.

Xyene: I feel this is less easy to screw up, since the calling script already defines this environment variable for determining which image to use, so there's a clear 1:1 mapping here.

quantum5: I disagree with the assertion that the calling script is reliable. This script is supposed to be generic "test in the docker image", not linked to Travis as a platform. It is worth noting that you have to explicitly pass `-e ARCH="$ARCH"` to `docker run` for this to work.

PR #582, opened 2019-11-09
GO: allow `package main;`; closes #544

### LINKED ISSUE #544: Go "package main;" disallowed
Semicolons are optional in Go, but our package validator fails if the name is followed by a semicolon.

kiritofeng:


dmoj/executors/GO.py
```diff
@@ -5,6 +5,7 @@
 
 reinline_comment = re.compile(br'//.*?(?=[\r\n])')
 recomment = re.compile(br'/\*.*?\*/', re.DOTALL)
+repackage = re.compile(br'\s*package\s+main\s*;?\b')
```
quantum5: Why do you have a `\b` in the end? It's pointless, for if the character that follows is `\w`, it will always match by not matching `;`. If the character that follows is not `\w`, it can always match by making both `\s` and `;` not match anything and use `\b` as boundary between `main` and that character.

So essentially, you have `\s*package\s+main\b`.

PR #589, opened 2019-11-16
Hold GIL when calling pt_process::spawn; fixes #588

### LINKED ISSUE #588: submission `pid` can be zero causing the judge to kill itself on a requested abort
If a user aborts a submission before the pid is set [here](https://github.com/DMOJ/judge-server/blob/67e856d0dd7f284d9f825ea67df7030380976b1a/dmoj/cptbox/ptproc.cpp#L84), then `os.killpg(0, signal.SIGKILL)` kills the judge itself.

According to @Xyene, the solution would be to remove `with nogil:` [here](https://github.com/DMOJ/judge-server/blob/c4b5e52665dfb6100dfa2de7429a05fd1a72ae7f/dmoj/cptbox/_cptbox.pyx#L433-L434), though this change needs to be tested before it is removed.

Ninjaclasher:
Fixes #588.

More testing will be done before this is marked ready for review.

### DIFF
--- a/dmoj/cptbox/_cptbox.pyx
+++ b/dmoj/cptbox/_cptbox.pyx
@@ -430,10 +430,8 @@ cdef class Process:
             if not self.debugger.is_exit(i):
                 config.syscall_whitelist[i] = self._syscall_whitelist[i]
 
-        with nogil:
-            if self.process.spawn(pt_child, &config):
-                with gil:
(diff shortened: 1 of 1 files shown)


quantum5:
Have fun with concurrency, which is *definitely* my favourite subject in university.

PR #594, opened 2019-12-08
RUST: strip binaries to lower memory overhead

Xyene:
/cc @blin00


blin00:
cc @goffrie :^)

goffrie:
never heard of "RUST"

PR #595, opened 2019-12-19
checker: Fix errors in `bridged` checker

kiritofeng:


dmoj/checkers/bridged.py
```diff
@@ -7,45 +7,28 @@
...
     if executor is None:
-        if 'files' not in checker_kwargs:
-            raise InternalError('no checker file(s) specified!')
-        if 'lang' not in checker_kwargs:
-            raise InternalError('language not specified for checker!')
-
-        filenames = list(map(lambda x: os.path.join(get_problem_root(problem_id), x), checker_kwargs['files']))
-        lang = checker_kwargs['lang']
-        executor = compile_with_auxiliary_files(filenames, lang, checker_kwargs['compiler_time_limit'])
+        filenames = list(map(lambda x: os.path.join(get_problem_root(problem_id), x), files))
```
quantum5: As a general rule, can we stop with the `list(map(...))` nonsense and just use comprehensions?

kiritofeng:
@Xyene requested that I drop `checker_kwargs`, and as the checker was previously broken, it wasn't used much on prod so there's not much this breaks w.r.t. backwards compatibility

PR #609, opened 2020-01-13
Fix `compile_with_auxiliary_files`

kiritofeng:
* Allows for passing of flags to generators and bridged checkers
* Allows for generators and bridged checkers to be not cached
* Fixes regression with bridged checker introduced in 0c08757
* Convert `dmoj.utils.aux_files` to Unix line endings

### DIFF
--- a/dmoj/checkers/bridged.py
+++ b/dmoj/checkers/bridged.py
@@ -8,17 +8,15 @@ from dmoj.result import CheckerResult
 from dmoj.utils.helper_files import compile_with_auxiliary_files, mktemp
 from dmoj.utils.unicode import utf8text
 
-executor = None
 
+def get_executor(problem_id, files, flags, lang, compiler_time_limit, should_cache):
+    if isinstance(files, str):
+        filenames = [files]
+    elif isinstance(files.unwrap(), list):
+        filenames = list(files.unwrap())
 
-def get_executor(files, lang, compiler_time_limit, problem_id):
-    global executor
-
-    if executor is None:
-        if not isinstance(files, list):
-            files = [files]
-        filenames = [os.path.join(get_problem_root(problem_id), f) for f in files]
-        executor = compile_with_auxiliary_files(filenames, lang, compiler_time_limit)
+    filenames = [os.path.join(get_problem_root(problem_id), f) for f in filenames]
+    executor = compile_with_auxiliary_files(filenames, flags, lang, compiler_time_limit, should_cache)
 
     return executor
 
@@ -34,11 +32,13 @@ def check(
     memory_limit=env['generator_memory_limit'],
     compiler_time_limit=env['generator_compiler_limit'],
     feedback=True,
+    flags=[],
+    cached=True,
     type='default',
     point_value=None,
     **kwargs
 ) -> CheckerResult:
-    executor = get_executor(files, lang, compiler_time_limit, problem_id)
+    executor = get_executor(problem_id, files, flags, lang, compiler_time_limit, cached)
 
     if type not in contrib_modules:
         raise InternalError('%s is not a valid return code parser' % type)
--- a/dmoj/generator.py
+++ b/dmoj/generator.py
@@ -4,6 +4,6 @@ from dmoj.utils.helper_files import compile_with_auxiliary_files
 
 
 class GeneratorManager:
-    def get_generator(self, filenames, flags, lang=None, compiler_time_limit=None):
+    def get_generator(self, filenames, flags, lang=None, compiler_time_limit=None, should_cache=True):
         filenames = list(map(os.path.abspath, filenames))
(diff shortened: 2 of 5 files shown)


dmoj/checkers/bridged.py
```diff
@@ -10,23 +10,26 @@
 executor = None
 
 
-def get_executor(files, lang, compiler_time_limit, problem_id):
+def get_executor(files, lang, flags, cached, compiler_time_limit, problem_id):
```
Xyene: `should_cache` or `is_cached`

dmoj/checkers/bridged.py
```diff
@@ -10,23 +10,26 @@
...
 
     if executor is None:
-        if not isinstance(files, list):
-            files = [files]
-        filenames = [os.path.join(get_problem_root(problem_id), f) for f in files]
-        executor = compile_with_auxiliary_files(filenames, lang, compiler_time_limit)
+        if isinstance(files, str):
+            filenames = [files]
+        elif isinstance(files.unwrap(), list):
+            filenames = list(files.unwrap())
```
Xyene: Wait why

dmoj/generator.py
```diff
@@ -4,6 +4,6 @@
 
 
 class GeneratorManager:
-    def get_generator(self, filenames, flags, lang=None, compiler_time_limit=None):
+    def get_generator(self, filenames, flags, lang=None, compiler_time_limit=None, cached=True):
         filenames = list(map(os.path.abspath, filenames))
-        return compile_with_auxiliary_files(filenames, lang, compiler_time_limit)
+        return compile_with_auxiliary_files(filenames, flags, lang, compiler_time_limit, cached)
```
Xyene: Try to keep parameters that are passed together in the same relative location. bridged.py sends in `flags, cached`, but this sends `flags, ..., cached`

dmoj/problem.py
```diff
@@ -266,14 +267,15 @@ def _run_generator(self, gen, args=None):
             time_limit = gen.time_limit or time_limit
             memory_limit = gen.memory_limit or memory_limit
             compiler_time_limit = gen.compiler_time_limit or compiler_time_limit
+            cached = gen.get('cached', True)
```
Xyene: When would you want something not to be cached?

dmoj/utils/aux_files.py
```diff
@@ -1,57 +1,62 @@
...
+
+    fs = executor.fs + [tempfile.gettempdir()]
+    executor = type('Executor', (executor,), {'fs': fs})
+
+    if issubclass(executor, CompiledExecutor):
+        executor = type('Executor', (executor,), {'compiler_time_limit': compiler_time_limit})
+
+    if hasattr(executor, 'flags'):
+        # We shouldn't be mutating the base class flags.
+        # See <https://github.com/DMOJ/judge-server/issues/174>.
```
Xyene: Was this comment... copied from somewhere? 174 was ages ago

kiritofeng: You deleted it when removing wbox

dmoj/checkers/bridged.py
kiritofeng: Same behaviour as generators

Xyene: Unless I'm missing something this is just `filenames = files`?

dmoj/problem.py
kiritofeng: If you are testing the generator caching it seems wasteful.

Also because there's currently a bug with the caching system, and this is a bit of a hotfix until I can determine the real cause of the bug

dmoj/checkers/bridged.py
kiritofeng: No `filenames` is a `ConfigNode`

dmoj/generator.py
kiritofeng: Now passed in same order as `compile_with_auxiliary_files`.

dmoj/utils/helper_files.py
```diff
@@ -13,7 +13,7 @@ def mktemp(data):
     return tmp
 
 
-def compile_with_auxiliary_files(filenames, lang=None, compiler_time_limit=None):
+def compile_with_auxiliary_files(filenames, flags=[], lang=None, compiler_time_limit=None, should_cached=True):
```
Xyene, line 16: "should_cached" sounds very weird

Xyene: It's also the wrong name as per https://github.com/DMOJ/judge-server/pull/609/files#diff-57b85c3e8735c00d8b1353b8903360ffR281

Xyene: Please test this code

dmoj/problem.py
kiritofeng: Update: the "bug" is not a result of the caching system, but the way it was being called.

PR #400, opened 2019-03-16
Allow submissions to use a different subset of testcases

kiritofeng:

PR #610, opened 2020-01-21
Move test case resolution into grader

kiritofeng:
As decided in #400, test case resolution is done by the grader, not the problem, to allow for the cases to change in response to the metadata.


kiritofeng:
@Xyene has volunteered to make this his problem.

PR #611, opened 2020-01-24
checker: Move `linecount` line-by-line feedback to `extended_feedback`

### LINKED ISSUE #407: Feedback from `linecount` checker overwrites IR feedback
After #398, we run `linecount` even if the submission gets an `IR` verdict. This results in the per-case feedback overwriting the error message's feedback.

Possible solutions include
1. Moving the per-case feedback into `extended_feedback`
2. Creating a new field to handle RTE/IR feedback

kiritofeng:
Closes #407

### DIFF
--- a/dmoj/checkers/linecount.py
+++ b/dmoj/checkers/linecount.py
@@ -28,7 +28,8 @@ def check(
             count += 1
 
     return CheckerResult(
-        count == len(judge_lines), point_value * (1.0 * count / len(judge_lines)), ''.join(cases) if feedback else ""
+        count == len(judge_lines), point_value * count / len(judge_lines),
+        extended_feedback='Case Feedback:\n' + ''.join(cases) if feedback else ''
     )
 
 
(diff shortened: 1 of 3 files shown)


quantum5:
Can you fix the spelling error in the commit message?

dmoj/checkers/linecount.py
```diff
@@ -28,7 +28,8 @@ def check(
             count += 1
 
     return CheckerResult(
-        count == len(judge_lines), point_value * (1.0 * count / len(judge_lines)), ''.join(cases) if feedback else ""
+        count == len(judge_lines), point_value * (1.0 * count / len(judge_lines)),
+        extended_feedback='Case Feedback:\n' + '\n'.join(cases) if feedback else ''
```
quantum5, line 32: `''.join` might be better so it's not one giant column going all the way down.

@Xyene any thoughts?

fluix-dev: Not Xyene, but if my memory serves me right then `''.join` is what MCPT does and it looks quite nice. I think the giant column wouldn't look nice either.

quantum5 (approved):

Xyene (approved):
Thanks!

PR #617, opened 2020-01-28
java: handle new failure mode for JVM MLE

Xyene:
"Newer" (post-Java 8) JVMs regressed a bit in terms of handling out-of-memory situations during
initialization, whereby they now dump a crash log rather than exiting with
java.lang.OutOfMemoryError. Handle this case so that we don't erroneously emit internal errors.


Xyene:
IR; for parity with `java.lang.OutOfMemoryError`, which is not marked as MLE.

PR #632, opened 2020-03-21
judgeenv: memoize problem root lookups

Xyene:
Follow-up to 59b98ea, allowing existing checkers to see speedups as
well, and reduce reliance on VM caching setup.


Xyene:
Wrong branch...

PR #633, opened 2020-03-21
judgeenv: memoize problem root lookups

Xyene:
Follow-up to 59b98ea, allowing existing checkers to see speedups as
well, and reduce reliance on VM caching setup.

### DIFF
--- a/dmoj/judgeenv.py
+++ b/dmoj/judgeenv.py
@@ -1,7 +1,7 @@
 import argparse
 import os
 import ssl
-from typing import List, Set
+from typing import Dict, List, Set
 
 import yaml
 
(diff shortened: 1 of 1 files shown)


dmoj/judgeenv.py
```diff
@@ -179,12 +179,20 @@ def get_path(x, y):
...
-    for dir in get_problem_roots():
-        path = os.path.join(dir, pid)
-        if os.path.exists(path):
-            return path
-    return None
+_problem_root_cache: Dict[str, str] = {}
+
+
+def get_problem_root(problem_id):
+    global _problem_root_cache
```
quantum5: Why do you need `global`?

PR #642, opened 2020-04-04
Limit number of digits interactor accepts; fixes #606

### LINKED ISSUE #606: Interactive graders can cause judge to hang due to large integers.
The `readint()` function in the interactive grader does not limit the number of digits in the integer.

When parsing an integer with `int()`, a very large (>10^7 digits) integer can cause the judge to hang. See [this submission](https://dmoj.ca/submission/1805481). This only affects interactive problems with longer time limits because 2 seconds is not currently enough to output all the digits of such an integer in an unbuffered problem.

Just to note that this also affects all custom checkers that use `int()` without limiting the number of digits, including those with time limits as low as 1 second.

A reasonable limit could be 10^5 digits, which is more then reasonable for all problems and [does not cause the judge to hang](https://dmoj.ca/submission/1805503). 

Another possible solution is to move to sandboxed interactors with the bridged grader.

Xyene:


Xyene:
However, it does not represent the maximum token length as returned by `readtoken`, so that would be misleading.

PR #652, opened 2020-04-11
Skip self-test for Java and ASM executors

kiritofeng:


dmoj/executors/asm_executor.py
```diff
@@ -112,7 +112,7 @@ def initialize(cls):
             return False
         if any(not os.path.isfile(i) for i in cls.crt_pre) or any(not os.path.isfile(i) for i in cls.crt_post):
             return False
-        return cls.run_self_test()
+        return skip_self_test or cls.run_self_test()
```
Xyene, line 115:
```suggestion
        # TODO(kirito): this code is also copied in java_executor.py, but judge should be refactored to call
        # `run_self_test` outside of `initialize`.
        return skip_self_test or cls.run_self_test()
```

PR #657, opened 2020-04-29
Correct bytes/str error in bridged checker

kiritofeng:


quantum5:
It's not immediately obvious to me why stderr should be bytes.

kiritofeng:
If `stderr` is a string, you will end up with the following error:
```
 File "/judge/dmoj/judge.py", line 188, in grading_cleanup_wrapper
   self._block_and_grade(problem, language, source, short_circuit, report=report)
 File "/judge/dmoj/judge.py", line 116, in _block_and_grade
   for result in self.grade_cases(
 File "/judge/dmoj/judge.py", line 235, in grade_cases
   result = grader.grade(case)
 File "/judge/dmoj/graders/standard.py", line 28, in grade
   check = self.check_result(case, result)
 File "/judge/dmoj/graders/standard.py", line 66, in check_result
   check = checker(
 File "/judge/dmoj/checkers/bridged.py", line 59, in check
   return contrib_modules[type].ContribModule.parse_return_code(
 File "/judge/dmoj/contrib/default.py", line 18, in parse_return_code
   parse_helper_file_error(proc, executor, name, stderr, time_limit, memory_limit)
 File "/judge/dmoj/utils/helper_files.py", line 80, in parse_helper_file_error
   feedback = Result.get_feedback_str(stderr, proc, executor)
 File "/judge/dmoj/result.py", line 67, in get_feedback_str
   feedback = binary.parse_feedback_from_stderr(error, process)
 File "/judge/dmoj/executors/gcc_executor.py", line 87, in parse_feedback_from_stderr
   match = deque(recppexc.finditer(stderr), maxlen=1)
TypeError: cannot use a bytes pattern on a string-like object
```

PR #658, opened 2020-05-02
Add executor for zig

kipply:
Not sure what else CI runs, but I ran `.docker.test.py` and AC
﻿
image

### DIFF
--- a/.docker.test.py
+++ b/.docker.test.py
@@ -26,6 +26,7 @@ elif arch == 'aarch64':
         'PY3',
         'SED',
         'TEXT',
+        'ZIG'
     }
     ALLOW_FAIL = set()
 else:
--- a/.docker.test.py
+++ b/.docker.test.py
@@ -26,7 +26,6 @@ elif arch == 'aarch64':
         'PY3',
         'SED',
         'TEXT',
-        'ZIG',
     }
     ALLOW_FAIL = set()
 else:
--- a/.docker.test.py
+++ b/.docker.test.py
@@ -26,7 +26,7 @@ elif arch == 'aarch64':
         'PY3',
         'SED',
         'TEXT',
-        'ZIG'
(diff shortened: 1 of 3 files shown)


.docker.test.py
```diff
@@ -26,6 +26,7 @@
         'PY3',
         'SED',
         'TEXT',
+        'ZIG'
```
Xyene, line 29:
```suggestion
        'ZIG',
```

(to prevent double-line diff the next time a language is added)

dmoj/executors/ZIG.py
```diff
@@ -0,0 +1,27 @@
...
+class Executor(CompiledExecutor):
+    ext = 'zig'
+    name = 'ZIG'
+    command = 'zig'
+    test_program = '''
+const std = @import("std");
+
+pub fn main() !void {
+    const stdout = std.io.getStdOut().outStream();
+    try stdout.print("echo: Hello, {}!\\n", .{"World"});
```
Xyene, line 13: If Zig has a way to read all lines of stdin, it should be used here rather than hardcoding the output of the self test.

kipply: added, though this does depend on trailing new lines I believe

.docker.test.py
```diff
@@ -26,6 +26,7 @@
         'PY3',
         'SED',
         'TEXT',
+        'ZIG',
```
Xyene, line 29: aarch64 only tests tier 1, and Zig is in tier 3, so this will fail all the time.

kipply: oh do I need to add it anywhere else?

Xyene: No, it should be picked up by `get_available` in the x86_64 branch above.

Xyene:
Thanks!

PR #661, opened 2020-05-05
Turn on --release-safe mode for Zig

fengb:
This is equivalent to -O2 in C compilers


Xyene:
FWIW, I've rolled over the judges with this + a fix for the newline issue, and rejudged existing Zig submissions.

PR #663, opened 2020-05-09
cptbox: validate null file paths

Xyene:
Otherwise, we can end up failing ptrace PEEKDATA, and return a garbage
path that only fails the sandbox check by accident.

### DIFF
--- a/dmoj/cptbox/ptdebug.cpp
+++ b/dmoj/cptbox/ptdebug.cpp
@@ -12,12 +12,6 @@
 
 pt_debugger::pt_debugger() : on_return_callback(NULL) {}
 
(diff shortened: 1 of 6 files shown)


dmoj/cptbox/ptdebug.cpp
```diff
@@ -118,6 +118,10 @@ ssize_t __attribute__((weak)) process_vm_readv(
 #endif
 
 char *pt_debugger::readstr(unsigned long addr, size_t max_size) {
+    if (!addr) {
+      return nullptr;
```
quantum5:
```suggestion
        return nullptr;
```

PR #669, opened 2020-05-19
cptbox: add better logging for child spawn failures

Xyene:


dmoj/cptbox/ptdebug_arm64.cpp
```diff
@@ -98,10 +98,14 @@ int pt_debugger_arm64::getpid_syscall() {
 }
 
 pt_debugger_arm64::pt_debugger_arm64() {
+#if PTBOX_SECCOMP
+    execve_id = 221;
```
quantum5, line 102: What's the impact of this? Is situation in the comments below not applicable for seccomp?

Xyene: seccomp has no exit event.

PR #673, opened 2020-05-25
Spawn a separate judge process to deal with each submission

### LINKED ISSUE #650: Investigate arena fragmentation issues
This ticket is a catch-all so I don't forget that this is an ongoing problem.

Xyene:
Fixes #650.

### DIFF
--- a/dmoj/cptbox/_cptbox.pyx
+++ b/dmoj/cptbox/_cptbox.pyx
@@ -429,6 +429,8 @@ cdef class Process:
             # a simple assembly program could terminate without ever trapping.
             if not self.debugger.is_exit(i):
                 config.syscall_whitelist[i] = self._syscall_whitelist[i]
+            else:
+                config.syscall_whitelist[i] = False
 
         if self.process.spawn(pt_child, &config):
             raise RuntimeError('failed to spawn child')
--- a/dmoj/cptbox/helper.cpp
+++ b/dmoj/cptbox/helper.cpp
@@ -4,6 +4,7 @@
 #include <dirent.h>
 #include <errno.h>
 #include <signal.h>
+#include <string.h>
 #include <unistd.h>
 #include <stdlib.h>
 #include <sys/resource.h>
@@ -156,27 +157,31 @@ int cptbox_child_run(const struct child_config *config) {
     if (config->trace_syscalls) {
         scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_TRACE(0));
         if (!ctx) {
-          perror("seccomp_init");
-          return 203;
+            fprintf(stderr, "Failed to initialize seccomp context!");
+            return 203;
         }
 
         unsigned int child_arch = get_seccomp_arch(config->debugger_type);
 
+        int rc;
         if (seccomp_arch_exist(ctx, child_arch) == -EEXIST &&
-            seccomp_arch_add(ctx, child_arch) != 0) {
-            perror("seccomp_arch_add");
+            (rc = seccomp_arch_add(ctx, child_arch)) != 0) {
+            fprintf(stderr, "seccomp_arch_add: %s\n", strerror(-rc));
             return 203;
         }
 
         for (int syscall = 0; syscall < MAX_SYSCALL; syscall++) {
             if (config->syscall_whitelist[syscall]) {
-                seccomp_rule_add(ctx, SCMP_ACT_ALLOW, syscall, 0);
+                if (rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, syscall, 0)) {
+                    fprintf(stderr, "seccomp_rule_add(..., %d): %s\n", syscall, strerror(-rc));
+                    // This failure is not fatal, it'll just cause the syscall to trap anyway.
+                }
             }
         }
 
-        if (seccomp_load(ctx)) {
-          perror("seccomp_load");
-          return 203;
+        if (rc = seccomp_load(ctx)) {
+            fprintf(stderr, "seccomp_load: %s\n", strerror(-rc));
+            return 203;
         }
 
         seccomp_release(ctx);
--- a/dmoj/judge.py
+++ b/dmoj/judge.py
@@ -3,16 +3,21 @@ import logging
 import os
 import signal
 import sys
+import multiprocessing
 import threading
 import traceback
+from enum import Enum
 from http.server import HTTPServer
+from itertools import groupby
+from operator import itemgetter
+from typing import Any, Callable, Dict, Generator, List, NamedTuple, Optional, Tuple, Union
 
-from dmoj import graders, packet
+from dmoj import packet
 from dmoj.control import JudgeControlRequestHandler
 from dmoj.error import CompileError
 from dmoj.judgeenv import clear_problem_dirs_cache, env, get_supported_problems, startup_warnings
 from dmoj.monitor import Monitor
-from dmoj.problem import BatchedTestCase, Problem
+from dmoj.problem import BatchedTestCase, Problem, TestCase
 from dmoj.result import Result
 from dmoj.utils.ansi import ansi_style, print_ansi, strip_ansi
 from dmoj.utils.unicode import unicode_stdout_stderr, utf8bytes, utf8text
@@ -30,34 +35,54 @@ except ImportError:
         pass
 
 
-class BatchBegin:
-    pass
+class IPC(Enum):
+    HELLO = 'HELLO'
+    BYE = 'BYE'
+    COMPILE_ERROR = 'COMPILE-ERROR'
+    COMPILE_MESSAGE = 'COMPILE-MESSAGE'
+    RESULT = 'RESULT'
+    BATCH_BEGIN = 'BATCH-BEGIN'
+    BATCH_END = 'BATCH-END'
+    GRADING_BEGIN = 'GRADING-BEGIN'
+    GRADING_END = 'GRADING-END'
+    GRADING_ABORTED = 'GRADING-ABORTED'
+    UNHANDLED_EXCEPTION = 'UNHANDLED-EXCEPTION'
+    REQUEST_ABORT = 'REQUEST-ABORT'
 
 
-class BatchEnd:
-    pass
-
-
-class TerminateGrading(Exception):
-    pass
+IPC_TEARDOWN_TIMEOUT = 5  # seconds
 
 
 logger = logging.getLogger(__name__)
 
 
+Submission = NamedTuple(
+    'Submission',
+    [
+        ('id', int),
+        ('problem_id', str),
+        ('language', str),
+        ('source', str),
+        ('time_limit', float),
+        ('memory_limit', int),
+        ('short_circuit', bool),
+        ('meta', Dict),
+    ],
+)
+
+
 class Judge:
-    def __init__(self):
-        self.current_grader = None
-        self.current_submission_id = None
-        self.current_submission_thread = None
-        self._terminate_grading = False
-        self.packet_manager = None
+    def __init__(self, packet_manager: packet.PacketManager) -> None:
+        self.packet_manager = packet_manager
+        # FIXME(tbrindus): marked as Any since PacketManager likes querying current_submission.id directly.
+        self.current_submission: Any = None
+        self.current_judge_worker: Optional[JudgeWorker] = None
 
         self.updater_exit = False
         self.updater_signal = threading.Event()
         self.updater = threading.Thread(target=self._updater_thread)
 
-    def _updater_thread(self):
+    def _updater_thread(self) -> None:
         log = logging.getLogger('dmoj.updater')
         while True:
             self.updater_signal.wait()
@@ -67,9 +92,10 @@ class Judge:
 
             # Prevent problem updates while grading.
             # Capture the value so it can't change.
-            thread = self.current_submission_thread
-            if thread:
-                thread.join()
+            # FIXME(tbrindus): this is broken.
+            # thread = sub_judge_thread
+            # if thread:
+            #    thread.join()
 
             try:
                 clear_problem_dirs_cache()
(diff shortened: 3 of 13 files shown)


dmoj/packet.py
```diff
@@ -304,60 +308,83 @@ def test_case_status_packet(self, position: int, result: Result):
...
     def begin_grading_packet(self, is_pretested: bool):
         log.info('Begin grading: %d', self.judge.current_submission_id)
-        self._send_packet(
+        self.send_raw_packet(
             {'name': 'grading-begin', 'submission-id': self.judge.current_submission_id, 'pretested': is_pretested}
         )
 
     def grading_end_packet(self):
         log.info('End grading: %d', self.judge.current_submission_id)
-        self.fallback = 4
```
Xyene, line 331: TODO: remember to add this back somehow

dmoj/judge.py
```diff
@@ -161,6 +163,42 @@ def _block_and_grade(self, problem, language, source, short_circuit, report=prin
         else:
             self.packet_manager.grading_end_packet()
 
+    def _run_grading_wrapper_process(
```
quantum5, line 166: Maybe use the prctl here to suicide when the parent dies?

dmoj/judge.py
```diff
@@ -80,51 +83,37 @@ def update_problems(self):
...
-                    self.packet_manager.batch_end_packet()
-                    batch_counter += 1
-                    in_batch = False
-                else:
+            # TODO(tbrindus): in what ways can these functions fail? Do we properly throw some IE if they do?
+            self.current_judge_process = SubmissionJudgeProcess(
+                problem_id, language, source, time_limit, memory_limit, short_circuit, meta
+            )
+            in_batch = False
+            for (type, data) in self.current_judge_process.run():
```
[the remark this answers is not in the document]
quantum5: Also... what if the sub process gives us garbage data that's invalid? Like say, ending the grading twice? Do we just let the bridge go wtf and disconnect us?

dmoj/judge.py
```diff
@@ -175,41 +179,202 @@ def begin_grading(
...
+        if blocking and sub_judge_thread is not None:
+            sub_judge_thread.join()
+
+    def log_internal_error(self, exc=None, message=None):
+        if not message:
+            # If exc is exists, raise it so that sys.exc_info() is populated with its data
+            if exc:
+                try:
+                    raise exc
+                except:  # noqa: E722, we want to catch everything
```
quantum5, line 199: This existed before, but technically this is a race because the user could Ctrl-C just before `raise exc` and `sys.exc_info()` would be `KeyboardInterrupt`.

dmoj/judge.py
```diff
@@ -175,41 +179,202 @@ def begin_grading(
...
+        self._terminate_grading = True
+        if self.grader:
+            self.grader.terminate_grading()
+
+    def _run_grading_wrapper_process(self, conn):
+        try:
+            # TODO(tbrindus): set different process title?
+
+            # TODO(tbrindus): name all these threads for exception readability
+            threading.Thread(target=self._run_abort_request_handler_thread, args=(conn,)).start()
```
quantum5, line 311: Should this be a daemon thread?

dmoj/judge.py
```diff
@@ -175,41 +179,202 @@ def begin_grading(
...
+            # TODO(tbrindus): can this race and fail?
+            self.process_conn.send(None)
+            self.join()
+
+    # TODO(tbrindus): do we need some sort of barrier here? what happens if an abort is sent before
+    # _block_and_grade is called?
+    def _run_abort_request_handler_thread(self, conn):
+        # If this ever returns, it's the signal that we should terminate the process.
+        # TODO(tbrindus): if it fails... it means our parent died?
+        conn.recv()
```
quantum5, line 301: When grading finishes, it doesn't look like we send anything before closing the `Pipe`. Will this print a pointless error?

dmoj/judge.py
```diff
@@ -175,41 +179,202 @@ def begin_grading(
...
+                        batch_counter += 1
+                    else:
+                        # TODO(tbrindus): we can't pickle TestCase
+                        yield IPC_RESULT, (case_number, result)
+                        case_number += 1
+            except TerminateGrading:
+                yield IPC_GRADING_TERMINATED, None
+            else:
+                yield IPC_GRADING_END, None
+        except GeneratorExit:
```
quantum5, line 368: Why do we need to catch `GeneratorExit`?

dmoj/judge.py
Xyene: Yes, done.

dmoj/judge.py
Xyene: Actually this raced and never caused a problem because the process (and this thread) would exit before recv could raise. I've rejigged this all so there's a graceful teardown with two-sided ACK, even if the only message currently passed judge &rarr; sub-judge is abort.

dmoj/judge.py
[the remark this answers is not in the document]
Xyene: Applied suggestion, since I don't mind either way.

I don't think we should do validation ourselves, and let the bridge deal with it as we always have. In a way our pipe is almost a direct pipe to the network, after all.

dmoj/judge.py
Xyene: Fixed.

dmoj/judge.py
Xyene: Fixed.

dmoj/judge.py
Xyene: In case the consumer of the generator fails during `conn.send`, I had this crash on me.

dmoj/judge.py
```diff
@@ -80,51 +83,37 @@ def update_problems(self):
...
-                    batch_counter += 1
-                    in_batch = False
-                else:
+            # TODO(tbrindus): in what ways can these functions fail? Do we properly throw some IE if they do?
+            self.current_judge_process = SubmissionJudgeProcess(
+                problem_id, language, source, time_limit, memory_limit, short_circuit, meta
+            )
+            in_batch = False
+            for (type, data) in self.current_judge_process.run():
+                if type == IPC_COMPILE_ERROR:
```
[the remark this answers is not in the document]
Xyene: Didn't do this because of having to track `in_batch`, but I've refactored things so that the batch number (or None) is included as part of the RESULT IPC so this could be done easily.

dmoj/judge.py
Xyene: I've removed this since other changes made it unnecessary, I think.

Xyene:
Outstanding bug:

```
INFO 2020-05-25 23:59:15,630 57 packet Received disconnect request, shutting down...
WARNING 2020-05-25 23:59:15,631 3816 tracer Request the killing of process: 4035
INFO 2020-05-25 23:59:15,638 57 packet Submission aborted: 1449068
INFO 2020-05-25 23:59:15,641 57 packet Internal error: 1449068
Exception in thread Thread-648:
  File "/judge/dmoj/judge.py", line 144, in _run_grading_thread
    handler_func(report, *(data or tuple()))
  File "/judge/dmoj/judge.py", line 194, in _ipc_grading_terminated
    self.packet_manager.submission_terminated_packet()
  File "/judge/dmoj/packet.py", line 353, in submission_terminated_packet
    self._send_packet({'name': 'submission-terminated', 'submission-id': self.judge.current_submission_id})
  File "/judge/dmoj/packet.py", line 235, in _send_packet
    self.output.writelines((PacketManager.SIZE_PACK.pack(len(raw)), raw))
  File "/usr/lib/python3.8/socket.py", line 687, in write
    return self._sock.send(b)
  File "/usr/lib/python3.8/ssl.py", line 1175, in send
    return super().send(data, flags)
BrokenPipeError: [Errno 32] Broken pipe

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/judge/dmoj/judge.py", line 150, in _run_grading_thread
    self.log_internal_error()
  File "/judge/dmoj/judge.py", line 245, in log_internal_error
    self.packet_manager.internal_error_packet(strip_ansi(message))
  File "/judge/dmoj/packet.py", line 319, in internal_error_packet
    self._send_packet(
  File "/judge/dmoj/packet.py", line 235, in _send_packet
    self.output.writelines((PacketManager.SIZE_PACK.pack(len(raw)), raw))
  File "/usr/lib/python3.8/socket.py", line 687, in write
    return self._sock.send(b)
  File "/usr/lib/python3.8/ssl.py", line 1175, in send
    return super().send(data, flags)
BrokenPipeError: [Errno 32] Broken pipe
  File "/judge/dmoj/judge.py", line 552, in main
    judge.listen()
  File "/judge/dmoj/judge.py", line 214, in listen
    self.packet_manager.run()
  File "/judge/dmoj/packet.py", line 178, in run
    self._read_forever()
  File "/judge/dmoj/packet.py", line 151, in _read_forever
    self._receive_packet(self._read_single())
  File "/judge/dmoj/packet.py", line 267, in _receive_packet
    self.disconnect()
  File "/judge/dmoj/packet.py", line 183, in disconnect
    sys.exit(0)
SystemExit: 0

During handling of the above exception, another exception occurred:

  File "/usr/local/bin/dmoj", line 11, in <module>
    load_entry_point('dmoj', 'console_scripts', 'dmoj')()
  File "/judge/dmoj/judge.py", line 558, in main
    judge.murder()
  File "/judge/dmoj/judge.py", line 220, in murder
    self.terminate_grading()
  File "/judge/dmoj/judge.py", line 206, in terminate_grading
    self.current_judge_process.terminate_grading()
  File "/judge/dmoj/judge.py", line 305, in terminate_grading
    self.process_conn.send((IPC_REQUEST_ABORT, None))
  File "/usr/lib/python3.8/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/lib/python3.8/multiprocessing/connection.py", line 411, in _send_bytes
    self._send(header + buf)
  File "/usr/lib/python3.8/multiprocessing/connection.py", line 368, in _send
    n = write(self._handle, buf)
BrokenPipeError: [Errno 32] Broken pip
```

dmoj/cptbox/tracer.py
```diff
@@ -195,7 +195,9 @@ def wait(self):
         self._died.wait()
         if not self.was_initialized:
             if self.returncode == 203:
-                raise RuntimeError('failed to set up seccomp policy')
+                out = self.stdout.read() if self.stdout else '(nil)'
+                err = self.stderr.read() if self.stderr else '(nil)'
+                raise RuntimeError('failed to set up seccomp policy, stdout="%s", stderr="%s"' % (out, err))
```
Xyene, line 200: This and the .docker/entry change won't be included in the final PR, just need these for now to test.

dmoj/judge.py
```diff
@@ -175,107 +94,123 @@ def begin_grading(
...
+                IPC_BATCH_END: self._ipc_batch_end,
+                IPC_RESULT: self._ipc_result,
+                IPC_UNHANDLED_EXCEPTION: self._ipc_unhandled_exception,
+            }
+
+            for ipc_msg in self.current_judge_process.run():
+                type, data = ipc_msg
+
+                handler_func = ipc_handler_dispatch.get(type)
+                if handler_func is None:
```
quantum5, line 138: Maybe `except KeyError` would be more pythonic here.

dmoj/judge.py
```diff
@@ -175,107 +94,123 @@ def begin_grading(
...
+        codes = result.readable_codes()
+
+        # here be cancer
+        is_sc = result.result_flag & Result.SC
+        colored_codes = list(
+            map(lambda x: '#ansi[%s](%s|bold)' % ('--' if x == 'SC' else x, Result.COLORS_BYID[x]), codes)
+        )
+        colored_aux_codes = '{%s}' % ', '.join(colored_codes[1:]) if len(codes) > 1 else ''
+        colored_feedback = '(#ansi[%s](|underline)) ' % utf8text(result.feedback) if result.feedback else ''
+        case_info = (
```
quantum5, line 182: This is horrifying. You should definitely do `if not is_sc:`...

dmoj/judge.py
```diff
@@ -284,22 +219,220 @@ def listen(self):
...
+
+            type, data = ipc_msg
+            if type == IPC_REQUEST_ABORT:
+                self._abort_requested = True
+                if self.grader:
+                    self.grader.abort_grading()
+            else:
+                raise RuntimeError("unexpected IPC message %s" % str(ipc_msg))
+
+    def _run_grading_wrapper_process(self, conn):
```
quantum5, line 333: Maybe do `self.process_conn.close()` for good measure.

Xyene: Done.

dmoj/judge.py
Xyene: Done.

dmoj/judge.py
Xyene: Done.

dmoj/judge.py
```diff
@@ -174,108 +93,130 @@ def begin_grading(
...
+    def _ipc_result(self, report, batch_number, case_number, result):
+        codes = result.readable_codes()
 
-        # Strip ANSI from the message, since this might be a checker's CompileError
-        # ...we don't want to see the raw ANSI codes from GCC/Clang on the site.
-        # We could use format_ansi and send HTML to the site, but the site doesn't presently support HTML
-        # internal error formatting.
-        self.packet_manager.internal_error_packet(strip_ansi(message))
+        is_sc = result.result_flag & Result.SC
+        colored_codes = list(
```
quantum5, line 179: Maybe we should just use a list comprehension here instead of `list(map(...))`.

dmoj/judge.py
```diff
@@ -284,22 +225,237 @@ def listen(self):
...
+                yield IPC_BATCH_BEGIN, (batch_number,)
+
+                for ipc_type, data in self._grade_cases(
+                    grader,
+                    case.batched_cases,
+                    short_circuit=case.config['short_circuit'],
+                    is_short_circuiting=is_short_circuiting,
+                    in_batch=True,
+                ):
+                    yield ipc_type, data
```
quantum5, line 428: Should probably increment `case_number` here...

dmoj/judge.py
```diff
@@ -174,108 +93,130 @@ def begin_grading(
         report=print,
         blocking=False,
     ):
-        self.current_submission_id = id
+        self.current_submission_id = submission_id
+        sub_judge_thread = threading.Thread(
```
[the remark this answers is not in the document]
Xyene: You only wait on it if `blocking = True`, which is only the case for CLI submissions.

dmoj/judge.py
```diff
@@ -284,22 +223,238 @@ def listen(self):
...
+        if self.grader:
+            self.grader.abort_grading()
+
+    # TODO(tbrindus): do we need some sort of barrier here? what happens if an abort is sent before
+    # _block_and_grade is called?
+    def _run_judge_request_handler_thread(self, conn):
+        while True:
+            try:
+                ipc_msg = conn.recv()
+            except Exception:
```
quantum5, line 335: This feels ridiculously broad.

dmoj/judge.py
```diff
@@ -284,22 +223,238 @@ def listen(self):
...
+                print("Failed to send abort request, did it race?", file=sys.stderr)
+                traceback.print_exc()
+            self.join()
+
+    def _do_abort(self):
+        self._abort_requested = True
+        if self.grader:
+            self.grader.abort_grading()
+
+    # TODO(tbrindus): do we need some sort of barrier here? what happens if an abort is sent before
```
quantum5, line 329: You'd want to block on a grading start event before you read from the queue.

dmoj/judge.py
```diff
@@ -174,108 +93,128 @@ def begin_grading(
...
+
+            for ipc_msg in self.current_judge_process.run():
+                type, data = ipc_msg
+
+                try:
+                    handler_func = ipc_handler_dispatch[type]
+                except KeyError:
+                    raise RuntimeError("unexpected IPC message %s" % str(ipc_msg)) from None
+
+                handler_func(report, *(data or tuple()))
```
quantum5, line 143:
```suggestion
                handler_func(report, *(data or ()))
```
`tuple()` is a dumb way of saying `()`.

dmoj/judge.py
```diff
@@ -284,22 +223,238 @@ def listen(self):
...
+            raise
+        except:  # noqa: E722, we explicitly want to notify the parent of everything
+            # We can't pickle the whole traceback object, so just send the formatted exception.
+            exc = sys.exc_info()
+            message = ''.join(traceback.format_exception(*exc))
+            conn.send((IPC_UNHANDLED_EXCEPTION, (message,)))
+            conn.send(None)
+        finally:
+            if comm_thread is not None:
+                comm_thread.join(timeout=5)
```
quantum5, line 375: What happens if this timeout is exceeded?

dmoj/judge.py
Xyene: Done.

dmoj/judge.py
```diff
@@ -284,22 +223,238 @@ def listen(self):
...
+        else:
+            binary = self.grader.binary
+            if hasattr(binary, 'warning') and binary.warning is not None:
+                yield IPC_COMPILE_MESSAGE, (binary.warning,)
+
+        yield from self._grade_cases(self.grader, self.grader.cases(), short_circuit=self.short_circuit)
+
+    def _grade_cases(self, grader, cases, short_circuit=False, is_short_circuiting=False, in_batch=False, case_number=0):
+        if not in_batch:
+            yield IPC_GRADING_BEGIN, (self.grader.is_pretested,)
```
quantum5, line 404: I am beginning to think it might be better to yield these in `_block_and_grade_cases`. It's annoying that we don't yield `IPC_GRADING_END` when aborted, however.

Xyene: It was like that before, and IMO it was much worse to follow because you had those `BatchBegin`/`BatchEnd`/`TerminateGrading` sentinels.

dmoj/judge.py
Xyene: I don't like it either, but _so far_ I've seen this function return:

* `BrokenPipeError`
* `OSError`
* `EOFError`
* `ConnectionResetError`

I don't really want to guess at what else it can return, and none of these have a nice parent class.

dmoj/judge.py
Xyene: Done.

dmoj/judge.py
Xyene: The subprocess exits anyway without necessarily having done the dual-FIN exchange, and the main judge will get an error when trying to read from the pipe. But we can't really wait here forever, I think.

Xyene: Added logging.

dmoj/judge.py
Xyene: Rewrote all of this.

dmoj/judge.py
```diff
@@ -284,22 +226,235 @@ def listen(self):
...
+        self.packet_manager.internal_error_packet(strip_ansi(message))
+
+        # Logs can contain ANSI, and it'll display fine
+        print(message, file=sys.stderr)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        pass
```
quantum5, line 267: Is there even a point to this context manager if it doesn't do any cleanup? i.e. now, `with Judge() as judge:` is the same as `judge = Judge()`.

dmoj/judge.py
```diff
@@ -284,22 +226,235 @@ def listen(self):
...
+
+        self.process_conn.close()
+        self.process.join()
+
+    def stop(self):
+        if self.process:
+            # Might be None if run was never called, or failed.
+            try:
+                self.process.join(timeout=IPC_TEARDOWN_TIMEOUT)
+                if self.process.is_alive():
```
quantum5, line 317: Maybe this should go into finally... In case something bad happens, we still want to kill the process.

dmoj/judge.py
```diff
@@ -284,22 +226,235 @@ def listen(self):
...
+                    self.process.kill()
+                    logger.error("Judge IPC process is still alive after timeout, sending SIGKILL!")
+            except OSError:
+                logger.exception('Exception while waiting for judge IPC process to shut down, ignoring...')
+
+    def abort_grading(self):
```
quantum5, line 323: What if an abortion was received _before_ `self.process` is assigned?

dmoj/judge.py
```diff
@@ -80,202 +85,139 @@ def update_problems(self):
...
+        sub_judge_thread = threading.Thread(
+            target=self._run_grading_thread,
+            args=(submission_id, problem_id, language, source, time_limit, memory_limit, short_circuit, meta, report),
+            daemon=True,
+        )
+        sub_judge_thread.start()
 
-        def grading_cleanup_wrapper():
+        # Submission may have been killed already, but block if it hasn't been.
+        if blocking and sub_judge_thread is not None:
```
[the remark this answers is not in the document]
Xyene: Can't happen anymore.

dmoj/judge.py
```diff
@@ -80,202 +85,139 @@ def update_problems(self):
...
+                IPC_GRADING_END: self._ipc_grading_end,
+                IPC_GRADING_ABORTED: self._ipc_grading_aborted,
+                IPC_BATCH_BEGIN: self._ipc_batch_begin,
+                IPC_BATCH_END: self._ipc_batch_end,
+                IPC_RESULT: self._ipc_result,
+                IPC_UNHANDLED_EXCEPTION: self._ipc_unhandled_exception,
+            }
+
+            for ipc_msg in self.current_judge_process.run():
+                type, data = ipc_msg
```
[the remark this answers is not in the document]
Xyene: Just because then you need to re-pack as a tuple when raising the  `RuntimeError`

dmoj/judge.py
Xyene: Yeah, removed.

dmoj/judge.py
Xyene: Good idea.

dmoj/judge.py
Xyene: I don't even know why this is guarded by `self.process`... it doesn't have to be.

I've added an `ipc_ready_signal` event though, so that `begin_grading` doesn't return before the IPC process is created, so we remove the race where an abort arrives while the IPC process is spawning.

dmoj/judge.py
```diff
@@ -83,202 +88,162 @@ def update_problems(self):
...
-            try:
-                problem = Problem(problem_id, time_limit, memory_limit, meta)
-                self._block_and_grade(problem, language, source, short_circuit, report=report)
-            except Exception:
-                self.log_internal_error()
+            self.current_judge_process = SubmissionJudgeProcess(
+                submission_id, problem_id, language, source, time_limit, memory_limit, short_circuit, meta
+            )
+
+            ipc_ready_signal.set()
```
quantum5, line 153: Does this help? The actual process may not be created here, so still a race.

Xyene: I think by this point in time we *have* at least forked, so the pipe will block rather than fail, which is fine for us.

Xyene: Wrote a fancier startup where judge waits for subjudge to say "HELLO" via IPC before setting the signal.

dmoj/judge.py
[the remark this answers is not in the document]
Xyene: Done anyways.

dmoj/judge.py
```diff
@@ -289,12 +222,224 @@ def murder(self):
...
+                    if (result.result_flag & Result.WA) and (
+                        batch_number or not case.points or self.submission.short_circuit
+                    ):
+                        is_short_circuiting = True
+
+                yield IPC_RESULT, (batch_number, case_number, result)
+
+            if batch_number:
+                yield IPC_BATCH_END, (batch_number,)
+                is_short_circuiting &= self.submission.short_circuit
```
Ninjaclasher, line 409: I think `is_short_circuiting` shouldn't be unset if the submission failed on a 0-point case (Prod example: https://dmoj.ca/submission/2110231)

Ninjaclasher: (The current behaviour is different than the behaviour before this commit)

Xyene: Thanks, fixed, I believe. https://dmoj.ca/submission/2110247, https://github.com/DMOJ/judge-server/compare/43156ec1a8bad901f998df0df6edf949c4574135..2c0c2a2639bf56a20a261705d39243ed35bb63e2

dmoj/judge.py
```diff
@@ -30,34 +36,53 @@ def setproctitle(title):
...
 class Judge:
-    def __init__(self):
-        self.current_grader = None
-        self.current_submission_id = None
-        self.current_submission_thread = None
-        self._terminate_grading = False
-        self.packet_manager = None
+    def __init__(self, packet_manager: packet.PacketManager) -> None:
+        self.packet_manager = packet_manager
+        self.current_submission: Any = None
```
quantum5, line 78: Why not `Optional[Submission]`?

Xyene: Because then `packet` fails to type-check, since it accesses `judge.current_submission.id`, and I didn't want to `assert judge.current_submission` in _every single packet handler_.

dmoj/cptbox/_cptbox.pyx
```diff
@@ -429,6 +429,8 @@ cdef class Process:
             # a simple assembly program could terminate without ever trapping.
             if not self.debugger.is_exit(i):
                 config.syscall_whitelist[i] = self._syscall_whitelist[i]
+            else:
+                config.syscall_whitelist[i] = False
```
quantum5, line 433: Should ideally be separate PR.

dmoj/cptbox/helper.cpp
```diff
@@ -162,21 +163,25 @@ int cptbox_child_run(const struct child_config *config) {
...
+            (rc = seccomp_arch_add(ctx, child_arch)) != 0) {
+            fprintf(stderr, "seccomp_arch_add: %s\n", strerror(-rc));
             return 203;
         }
 
         for (int syscall = 0; syscall < MAX_SYSCALL; syscall++) {
             if (config->syscall_whitelist[syscall]) {
-                seccomp_rule_add(ctx, SCMP_ACT_ALLOW, syscall, 0);
+                if (rc = seccomp_rule_add(ctx, SCMP_ACT_ALLOW, syscall, 0)) {
+                    fprintf(stderr, "seccomp_rule_add(..., %d): %s\n", syscall, strerror(-rc));
```
quantum5, line 176: I suppose it makes no difference if this failed, since it will just end up trapping ptrace. Separate PR?

dmoj/judge.py
```diff
@@ -67,239 +93,377 @@ def _updater_thread(self):
...
+                try:
+                    ipc_type, data = judge_process_conn.recv()
+                except:  # noqa: E722, whatever happened, we have to abort now.
+                    logger.exception("Judge unexpectedly hung up!")
+                    self._do_abort()
+                    return
+
+                if ipc_type == IPC.BYE:
+                    return
+                if ipc_type == IPC.REQUEST_ABORT:
```
quantum5, line 350: Will this look better as `elif`?
```suggestion
                elif ipc_type == IPC.REQUEST_ABORT:
```

dmoj/problem.py
```diff
@@ -159,6 +159,19 @@ def load_checker(self, name):
         self._checkers[name] = checker = load_module_from_file(os.path.join(self.root_dir, name))
         return checker
 
+    @property
+    def grader_class(self):
```
quantum5, line 163: How did you fix the mypy here?

dmoj/judge.py
Xyene: No opinion, so I made it an `elif`.

dmoj/cptbox/helper.cpp
Xyene: Added explanatory comment, but we should keep the log.

dmoj/cptbox/_cptbox.pyx
Xyene: Yep, done.

dmoj/problem.py
Xyene: My secret move: I tacked on a `# FIXME(tbrindus)` with `Any` :slightly_smiling_face:

dmoj/judge.py
Xyene: Opted against this in the end so that we could do some best-effort reporting to show what went wrong. Any active submission will be prctl'd, though.

PR #676, opened 2020-05-26
Change packet log level to debug

Xyene:


dmoj/cptbox/syscalls/generate_linux.py
```diff
@@ -84,13 +84,16 @@
         names.add(line.split()[1])
 
 with open('../syscalls.pyi', 'w') as interface:
-    print('''\
+    print(
+        '''\
```
quantum5, line 88: Is this actually better?

PR #683, opened 2020-05-31
cptbox: ensure shocker thread exits on OLE

Xyene:


dmoj/cptbox/tracer.py
```diff
@@ -335,7 +336,7 @@ def _shocker_thread(self):
         wake_signal = signal.SIGSTOP if 'freebsd' in sys.platform else signal.SIGWINCH
         self._spawned_or_errored.wait()
 
-        while not self._exited:
+        while not self._died.is_set():
```
quantum5, line 339: If we are using a conditional variable, why use it as a flag? Why not just wait on it with limit of one second instead of `time.sleep` down there, and break if the wait succeeded?

Xyene: It slightly changes the behavior (no more `time.sleep(0.01)`), but I don't think that's meaningful. Done.

PR #695, opened 2020-06-03
Use an action queue to prevent updating problems mid-submission

### LINKED ISSUE #671: Problem updates can happen while grading
There is code to ensure that an _update_ doesn't start mid-submission, but by oversight it doesn't prevent a _submission_ starting mid-update. This is problematic because an update currently takes ~10s of rather high CPU usage, which can throw off the timings of a submission that starts mid-update.

Xyene:
Closes #671.

### DIFF
--- a/dmoj/judge.py
+++ b/dmoj/judge.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 import logging
 import os
+import queue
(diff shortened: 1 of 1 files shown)


Xyene:
This is a bad solution.

PR #700, opened 2020-06-13
Kill the judging submission first with a keyboard interrupt on CLI

Ninjaclasher:
This is more of a QOL improvement. Currently, when there is a long running submission (e.g a submission that TLEs on every case), there is no way to abort the submission. Pressing `Ctrl-C` would kill the CLI, potentially losing any commands that were previously typed and had scrolled off the screen (since there is no persistent history). Thus, the only way would be to wait out the grading.

In this PR, instead of killing the CLI directly while a submission is grading, we first kill the currently grading submission if there is one.


Xyene:
On that note should also make `readline` not-optional and just include it in the requirements.

PR #702, opened 2020-06-13
Add persistent command history storage for CLI

Ninjaclasher:


dmoj/cli.py
```diff
@@ -82,6 +83,11 @@ def cli_main():
     executors.load_executors()
     contrib.load_contrib_modules()
 
+    try:
+        readline.read_history_file(judgeenv.cli_history_file)
```
quantum5, line 87: This magically expands the ~ in path?

Ninjaclasher: Should be fixed now. Thanks for pointing that out.

PR #705, opened 2020-06-17
cptbox: fix spurious seccomp initialization failures

Xyene:
:blood_sob:

### DIFF
--- a/dmoj/cptbox/helper.cpp
+++ b/dmoj/cptbox/helper.cpp
@@ -103,27 +103,6 @@ int cptbox_child_run(const struct child_config *config) {
         personality(config->personality);
 #endif
 
-    if (config->address_space)
-        setrlimit2(RLIMIT_AS, config->address_space);
-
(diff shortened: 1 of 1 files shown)


dmoj/cptbox/helper.cpp
```diff
@@ -188,6 +167,30 @@ int cptbox_child_run(const struct child_config *config) {
     }
 #endif
 
+    // All these limits should be dropped after initializing seccomp, since seccomp allocates
+    // memory, and if an arena isn't sufficiently free it could force seccomp into an OOM
+    // situation where we'd fail to initialize.
+    if (config->address_space)
+        setrlimit2(RLIMIT_AS, config->address_space);
```
quantum5, line 174: It appears that if the seccomp rules don't allow setrlimit, then every single one of these calls will trap in ptrace. Fortunately, cptbox allows anything before the first execve, so this will run, but I would prefer that it doesn't trap if possible.

PR #710, opened 2020-06-22
Allow extra stderr output from COCI and testlib checkers

kiritofeng:
(Tested against https://github.com/DMOJ/judge-testsuite/pull/58)

### DIFF
--- a/dmoj/contrib/coci.py
+++ b/dmoj/contrib/coci.py
@@ -7,7 +7,7 @@ from dmoj.result import CheckerResult
 
 class ContribModule(TestlibContribModule):
     name = 'coci'
-    repartial = re.compile(br'^partial ((\d+)\/(\d*[1-9]\d*))\n$')
+    repartial = re.compile(br'^partial ((\d+)\/(\d*[1-9]\d*))$', re.M)
 
     @classmethod
     def parse_return_code(cls, proc, executor, point_value, time_limit, memory_limit, feedback, name, stderr):
--- a/dmoj/contrib/testlib.py
+++ b/dmoj/contrib/testlib.py
@@ -14,7 +14,7 @@ class ContribModule(DefaultContribModule):
     PARTIAL = 7
 
     name = 'testlib'
-    repartial = re.compile(br'^points (\d+)\n$')
+    repartial = re.compile(br'^points (\d+)$', re.M)
 
     @classmethod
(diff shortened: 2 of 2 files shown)


dmoj/contrib/coci.py
```diff
@@ -1,26 +1,26 @@
-import re
```
quantum5, line 1: Why is the entire file changed?

dmoj/contrib/coci.py
```diff
@@ -1,26 +1,26 @@
...
+import re
+
+from dmoj.contrib.testlib import ContribModule as TestlibContribModule
+from dmoj.error import InternalError
+from dmoj.result import CheckerResult
+
+
+class ContribModule(TestlibContribModule):
+    name = 'coci'
+    repartial = re.compile(br'^partial ((\d+)\/(\d*[1-9]\d*))\n$', re.M)
```
quantum5, line 10: Do we need/wait the `\n`?

dmoj/contrib/testlib.py
```diff
@@ -14,7 +14,7 @@ class ContribModule(DefaultContribModule):
     PARTIAL = 7
 
     name = 'testlib'
-    repartial = re.compile(br'^points (\d+)\n$')
+    repartial = re.compile(br'^points (\d+)\n$', re.M)
```
quantum5, line 17: Do we need/wait the `\n`?

dmoj/contrib/coci.py
kiritofeng: Windows line ending nonsense.

dmoj/contrib/coci.py
kiritofeng: IMO we should maintain parity with testlib.h, whatever that may be...

Also I would recommend against removing entirely, as then we'd have to modify every checker that uses partials...

dmoj/contrib/testlib.py
kiritofeng: IMO we should maintain parity with testlib.h, whatever that may be...

Also I would recommend against removing entirely, as then we'd have to modify every checker that uses partials...

quantum5: No I mean the regex will match the same either way. Do we just want to not care if testlib outputs the trailling `\n` or not?

kiritofeng: Decided to drop the trailing `\n`.

PR #711, opened 2020-06-23
Remove message about aborting not deleting files

Ninjaclasher:
It seems that after #673, temporary files are deleted after a submission aborts. Since the submission is run on a separate process now, the GC will cleanup any temporary files. Thus, the message `Temporary files may not be deleted.` is only partially correct, since temporary files are now _always_ deleted.

This PR removes the message.


Ninjaclasher:
Seems like temporary files are deleted only when using CPython (they're not deleted on PyPy, for example).

PR #712, opened 2020-06-24
cptbox: remove native arch from seccomp when executing x86-on-x64

Xyene:


Xyene:
Superseded by https://github.com/DMOJ/judge-server/pull/713.

PR #714, opened 2020-06-26
More robust waiting on worker

Xyene:


dmoj/judge.py
```diff
@@ -282,9 +286,17 @@ def __init__(self, submission: Submission) -> None:
         child_conn.close()
 
     def communicate(self) -> Generator[Tuple[IPC, tuple], None, None]:
+        recv_timeout = max(60, int(2 * self.submission.time_limit))
```
Xyene, line 289: Known issue: if compiler TL is > 60 for some reason, this could cause problems. But I won't care about that case until it becomes a problem; loading the entire `Problem` object here just for that field feels very heavyweight.

PR #724, opened 2020-07-18
executors: tag submission tempdir with submission ID where available

Xyene:
This will help tracking down obscure bugs where processes linger despite
their directories having been long removed.

### DIFF
--- a/dmoj/graders/standard.py
+++ b/dmoj/graders/standard.py
@@ -114,6 +114,7 @@ class StandardGrader(BaseGrader):
         return executors[self.language].Executor(
             self.problem.id,
             self.source,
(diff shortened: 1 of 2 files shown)


dmoj/graders/standard.py
```diff
@@ -114,6 +114,7 @@ def _generate_binary(self):
         return executors[self.language].Executor(
             self.problem.id,
             self.source,
+            tempdir_tag=str(self.judge.submission.id),
```
kiritofeng, line 117: Should we also pass this to the signature grader?

Xyene: Maybe, but I don't like the design of this patch as-is anyway (need it to look into a JVM failure).

PR #727, opened 2020-07-29
judge: attempt to make abort logic less racey

Xyene:


Xyene:
This code is hard to prove the correctness of (is it even correct?), but it has been running in prod on Lletya/Iorwerth for over a month now, and the judge _not_ running it (Crwys) frequently reports errors on abort.

PR #728, opened 2020-08-02
Use problem globs to declare problem storage roots

### LINKED ISSUE #723: Watchdog fails if `problem_storage_root` is a string
Should check if `problem_dirs` is a string before https://github.com/DMOJ/judge-server/blob/master/dmoj/judgeenv.py#L173, otherwise we will be watching nonsense paths,

### LINKED ISSUE #772: Use globs for `problem_storage_root`
Declare problem storage roots using globs, replacing all 3 previous implementations.

For judge v2, this should be called `problem_storage_globs`, for compatibility.

For judge v3, this should replace the current implementation of `problem_storage_root`.

kiritofeng:
Closes #772. Also supersedes #1026.

### DIFF
--- a/dmoj/judgeenv.py
+++ b/dmoj/judgeenv.py
@@ -1,4 +1,5 @@
 import argparse
+import glob
 import logging
 import os
 import ssl
@@ -12,6 +13,7 @@ from dmoj.utils import pyyaml_patch  # noqa: F401, imported for side effect
 from dmoj.utils.unicode import utf8text
 
 problem_dirs = ()
+problem_globs = ()
 problem_watches = ()
 env = ConfigNode(
     defaults={
@@ -67,7 +69,7 @@ exclude_executors: Set[str] = set()
 
 
 def load_env(cli=False, testsuite=False):  # pragma: no cover
-    global problem_dirs, only_executors, exclude_executors, log_file, server_host, server_port, no_ansi, no_ansi_emu, skip_self_test, env, startup_warnings, no_watchdog, problem_regex, case_regex, api_listen, secure, no_cert_check, cert_store, problem_watches, cli_history_file, cli_command, log_level
+    global problem_dirs, problem_globs, get_problem_roots, only_executors, exclude_executors, log_file, server_host, server_port, no_ansi, no_ansi_emu, skip_self_test, env, startup_warnings, no_watchdog, problem_regex, case_regex, api_listen, secure, no_cert_check, cert_store, problem_watches, cli_history_file, cli_command, log_level
 
     if cli:
         description = 'Starts a shell for interfacing with a local judge instance.'
@@ -134,6 +136,7 @@ def load_env(cli=False, testsuite=False):  # pragma: no cover
     parser.add_argument('--skip-self-test', action='store_true', help='skip executor self-tests')
 
     if testsuite:
+        # TODO: replace with globs
         parser.add_argument('tests_dir', help='directory where tests are stored')
         parser.add_argument('problem_regex', help='when specified, only matched problems will be tested', nargs='?')
         parser.add_argument('case_regex', help='when specified, only matched cases will be tested', nargs='?')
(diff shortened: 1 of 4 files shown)


Xyene:
This largely looks good to me. Could you update the PR description to include a sample config?

I'd like the next version we tag to be 3.0.0, since we have a fair number of breaking changes. So, I think we should:

* merge this PR
* queue up a PR totally removing the old config format, but don't merge it until right before tagging 3.0.0 (so we don't wreck people installing from master)
* queue up a PR updating the docs

kiritofeng:
As an example,
```yaml
problem_storage_root:
- 2: /problems/ccc/nccc
- 1: /problems/apio
- 1: /problems/bbc
```
would become
```yaml
problem_storage_globs:
- /problems/ccc/nccc/*/**/
- /problems/apio/*/apio*/
- /problems/bbc/*/bbc*/
```

dmoj/judgeenv.py
```diff
@@ -177,44 +180,54 @@ def load_env(cli=False, testsuite=False):  # pragma: no cover
...
+
+    if problem_globs is None and problem_dirs is None:
+        if not testsuite:
+            raise SystemExit('problem_storage_globs not specified in "%s"; no problems available to grade' % model_file)
 
     if testsuite:
         if not os.path.isdir(args.tests_dir):
             raise SystemExit('Invalid tests directory')
-        problem_dirs = [args.tests_dir]
+        problem_globs = [os.path.join(args.tests_dir, '*')]
```
kiritofeng, line 230: Something we could do (when we're ready to replace `problem_storage_root` with `problem_storage_globs`) is to replace the test regexes with test globs

dmoj/judgeenv.py
```diff
@@ -177,44 +179,54 @@ def load_env(cli=False, testsuite=False):  # pragma: no cover
...
-                problem_watches.append(get_path(_root, recursive_root))
-        else:
-            problem_watches.append(get_path(_root, dir))
+        if isinstance(problem_globs, str):
+            problem_globs = [problem_globs]
+
+        problem_watches = problem_globs
+
+    elif env.problem_storage_root:
+        startup_warnings.append('problem_storage_root is deprecated, use problem_storage_globs instead')
```
Xyene, line 203:
```suggestion
        startup_warnings.append('`problem_storage_root` is deprecated, use `problem_storage_globs` instead')
```

dmoj/judgeenv.py
Xyene: This sounds like a fine idea to leave as a comment :)

dmoj/judgeenv.py
```diff
@@ -177,44 +179,54 @@ def load_env(cli=False, testsuite=False):  # pragma: no cover
...
+        get_problem_roots(warnings=True)
 
-    problem_watches = []
-    for dir in problem_dirs:
-        if isinstance(dir, ConfigNode):
-            for _, recursive_root in dir.iteritems():
-                problem_watches.append(get_path(_root, recursive_root))
-        else:
-            problem_watches.append(get_path(_root, dir))
+        if isinstance(problem_globs, str):
```
Xyene, line 197: We should not support this mode, and instead always require a list here.

dmoj/judgeenv.py
```diff
@@ -324,10 +349,11 @@ def get_supported_problems_and_mtimes():
     """
     problems = []
     for dir in get_problem_roots():
-        for problem in os.listdir(dir):
-            problem = utf8text(problem)
-            if os.access(os.path.join(dir, problem, 'init.yml'), os.R_OK):
-                problems.append((problem, os.path.getmtime(os.path.join(dir, problem))))
+        if os.path.isdir(dir):
```
Xyene, line 352: When can this happen?

kiritofeng: if you delete a folder that previously contained a problem, it can cause the scan to crash

Xyene: Let's invert this to read `if not ...: continue` and leave a comment.

Is this new behavior, or did it always crash this way and this is fixing a bug?

dmoj/judgeenv.py
```diff
@@ -177,44 +179,54 @@ def load_env(cli=False, testsuite=False):  # pragma: no cover
...
+        for dir in problem_dirs:
+            if isinstance(dir, ConfigNode):
+                for _, recursive_root in dir.iteritems():
+                    problem_watches.append(get_path(_root, recursive_root))
+            else:
+                problem_watches.append(get_path(_root, dir))
+
+    if problem_globs is None and problem_dirs is None:
+        if not testsuite:
+            raise SystemExit('problem_storage_globs not specified in "%s"; no problems available to grade' % model_file)
```
Xyene, line 224:
```suggestion
            raise SystemExit('`problem_storage_globs` not specified in "%s"; no problems available to grade' % model_file)
```

dmoj/judgeenv.py
kiritofeng: Just tested on the judges we run in prod, the bug exists there too

Xyene (approved):
LGTM, feel free to merge once lint passes.

PR #731, opened 2020-08-03
Add bridged grader for WCIPEG problems; fixes #729

### LINKED ISSUE #729: Add bridged grader for WCIPEG problems
TODO:
- [ ] Write WCIPEG contrib module
- [ ] Allow bridged grader to pass the input, output, and answer files in a permutation as specified by the configuration

kiritofeng:


dmoj/checkers/bridged.py
```diff
@@ -43,11 +44,12 @@ def check(
     if type not in contrib_modules:
         raise InternalError('%s is not a valid return code parser' % type)
 
+    args = args or contrib_modules[type].ContribModule.get_args_string()
+
     with mktemp(judge_input) as input_file, mktemp(process_output) as output_file, mktemp(judge_output) as judge_file:
+        args = args.format(input=input_file.name, output=output_file.name, answer=judge_file.name).split()
```
kiritofeng, line 50: Should use `shlex.split()`

dmoj/checkers/bridged.py
```diff
@@ -35,23 +36,27 @@ def check(
...
     executor = get_executor(problem_id, files, flags, lang, compiler_time_limit, cached)
 
     if type not in contrib_modules:
-        raise InternalError('%s is not a valid return code parser' % type)
+        raise InternalError('%s is not a valid contrib module' % type)
+
+    args = args or contrib_modules[type].ContribModule.get_args_string()
 
     with mktemp(judge_input) as input_file, mktemp(process_output) as output_file, mktemp(judge_output) as judge_file:
+        args = shlex.split(
```
quantum5, line 51: The fact you are using shlex is disturbing, but I also can't think of a better way, so whatever.

dmoj/graders/bridged.py
```diff
@@ -54,14 +55,17 @@ def _launch_process(self, case):
...
-        output = case.output_data()
+        judge_output = case.output_data()
         self._interactor_time_limit = (self.handler_data.preprocessing_time or 0) + self.problem.time_limit
         self._interactor_memory_limit = self.handler_data.memory_limit or env['generator_memory_limit']
+        args = self.handler_data.args or contrib_modules[self.contrib_type].ContribModule.get_args_string()
 
-        with mktemp(input) as input_file, mktemp(output) as output_file:
+        with mktemp(input) as input_file, mktemp(judge_output) as judge_file:
+            args = shlex.split(
+                args.format(input=shlex.quote(input_file.name), output='', answer=shlex.quote(judge_file.name),)
```
quantum5, line 65: Doesn't this result in a blank argument? Isn't that a problem?

kiritofeng: Fixed by splitting into an args string for checker and one for interactor.
The one for default interactor doesn't have `{output}`, but reading the `testlib.h` documentation, they expect an `output` file, so made the testlib contrib module compliant.

This breaks backwards compatibility with exactly 1 interactor in production, but by passing `args: '{input}, {answer}'` in its init.yml, it is resolved.

dmoj/graders/bridged.py
```diff
@@ -54,14 +59,21 @@ def _launch_process(self, case):
         os.close(submission_stdout_pipe)
 
     def _interact_with_process(self, case, result, input):
-        output = case.output_data()
+        judge_output = case.output_data()
         self._interactor_time_limit = (self.handler_data.preprocessing_time or 0) + self.problem.time_limit
         self._interactor_memory_limit = self.handler_data.memory_limit or env['generator_memory_limit']
+        args = self.handler_data.args or contrib_modules[self.contrib_type].ContribModule.get_interactor_args_string()
```
Xyene, line 65: `s/args/args_format_string/g` everywhere, because we already use `args` to mean a *list* when it comes to generators so this would be overloading the same key (albeit with a different parent).

dmoj/checkers/bridged.py
```diff
@@ -35,23 +36,25 @@ def check(
...
 
     if type not in contrib_modules:
-        raise InternalError('%s is not a valid return code parser' % type)
+        raise InternalError('%s is not a valid contrib module' % type)
+
+    args_format_string = args_format_string or contrib_modules[type].ContribModule.get_checker_args_string()
 
     with mktemp(judge_input) as input_file, mktemp(process_output) as output_file, mktemp(judge_output) as judge_file:
+        checker_args = shlex.split(args_format_string.format(
+            input=shlex.quote(input_file.name),
```
Xyene, line 52: `input_file`, etc.

dmoj/checkers/bridged.py
```diff
@@ -35,23 +36,25 @@ def check(
...
-    **kwargs
+    **kwargs,
 ) -> CheckerResult:
     executor = get_executor(problem_id, files, flags, lang, compiler_time_limit, cached)
 
     if type not in contrib_modules:
-        raise InternalError('%s is not a valid return code parser' % type)
+        raise InternalError('%s is not a valid contrib module' % type)
+
+    args_format_string = args_format_string or contrib_modules[type].ContribModule.get_checker_args_string()
```
Xyene, line 48: `get_checker_args_format_string`

dmoj/contrib/coci.py
```diff
@@ -9,6 +9,10 @@ class ContribModule(TestlibContribModule):
     name = 'coci'
     repartial = re.compile(br'^partial ((\d+)\/(\d*[1-9]\d*))$', re.M)
 
+    @classmethod
+    def get_interactor_args_string(cls):
```
Xyene, line 13: `get_interactor_format_args_string`

PR #737, opened 2020-08-08
contrib: Correct inaccuracies in PEG contrib module 

kiritofeng:
According to @t3nsor, the correct order is `<checker> <judge answer> <program output> <input> <source file>`. (I suspect we can get away with not passing the source file in 99% of cases)

Additionally, printing a ratio of floats is valid.

NOI '04 P1 (which I based this code off of) just happened to have poor choice in variable names...

### DIFF
--- a/dmoj/contrib/peg.py
+++ b/dmoj/contrib/peg.py
@@ -1,5 +1,3 @@
-import re
-
 from dmoj.contrib.default import ContribModule as DefaultContribModule
 from dmoj.result import CheckerResult
 from dmoj.utils.helper_files import parse_helper_file_error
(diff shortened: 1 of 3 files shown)


dmoj/contrib/peg.py
```diff
@@ -7,11 +7,11 @@
 
 class ContribModule(DefaultContribModule):
     name = 'peg'
-    repartial = re.compile(r'^(\d+)\n(\d+)$', re.M)
+    repartial = re.compile(r'^(\d+\.?\d*)\n(\d+\.?\d*)$', re.M)
```
Xyene, line 10: This doesn't really match all floats, e.g. ones using scientific notation. Might be best to just split the lines and `float(...)`.

kiritofeng: Changed to checking for 2 consecutive lines that have valid floats.

PR #738, opened 2020-08-08
Add flag to allow siggrader to not overwrite main function; fixes #730

### LINKED ISSUE #730: Add flag to allow siggrader to not overwrite main function
Some problems (e.g. CEOI '18 P6) use signature grading to emulate interaction, and so require the main to be left intact

kiritofeng:
Some problems (e.g. CEOI '18 P6) use signature grading to emulate interaction, and so require the main to be left intact.

Fixes #730.

### DIFF
--- a/dmoj/graders/signature.py
+++ b/dmoj/graders/signature.py
@@ -26,7 +26,7 @@ class SignatureGrader(StandardGrader):
 
             submission_prefix = '#include "%s"\n' % handler_data['header']
             if not handler_data.get('allow_main', False):
-                submission_prefix += '#define main main_%s\n' % str(uuid.uuid4()).replace('-', '')
+                submission_prefix += '#define main main_%s\n' % uuid.uuid4().hex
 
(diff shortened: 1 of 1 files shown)


dmoj/graders/signature.py
```diff
@@ -24,12 +24,9 @@ def _generate_binary(self):
...
 
-            # fmt: off
-            submission_prefix = (
-                '#include "%s"\n'
-                '#define main main_%s\n'
-            ) % (handler_data['header'], str(uuid.uuid4()).replace('-', ''))
-            # fmt: on
+            submission_prefix = '#include "%s"\n' % handler_data['header']
+            if not handler_data.get('allow_main', False):
+                submission_prefix += '#define main main_%s\n' % str(uuid.uuid4()).replace('-', '')
```
Ninjaclasher, line 29:
```suggestion
                submission_prefix += '#define main main_%s\n' % uuid.uuid4().hex
```
I think these should be equivalent, and grabbing the hex directly makes it more clear what we are trying to do.

PR #757, opened 2020-09-30
judge: disallow running as root

Xyene:
Letting this be an option is an invitation for people to use it without understanding the (major) repercussions, so let's just remove it.


quantum5 (approved):

TheAvidDev:
Do we care for explaining why?

PR #759, opened 2020-09-30
Fix `problem_watches` watching nonsense directories; #723

kiritofeng:
Problem globs is on hold due to time constraints, and this fix should occur regardless (as we decided v2 of judge should still support `problem_storage_root`).


Xyene (approved):
:+1:

Issue #697, opened 2020-06-06
`was_initialized` is always false if using `NullTracer`


Xyene:
This is expected since we never see the `execve` if we're not tracing to begin with, but unexpected because we should assume true in such a case. The Java executors use `NullTracer`, and can cause problems in code expecting `was_initialized` to always be sensible.

PR #764, opened 2020-10-03
Remove `NullTracer` implementation

Xyene:
It is no longer being used.


quantum5 (approved):

kiritofeng:
Does this close #697? If so, you should probably tag it in the commit.

Xyene:
It doesn't; this code is just not used. The #697 description is misleading.

PR #765, opened 2020-10-03
Disable codecov status

quantum5:
This eliminates the spam of errors that we don't actually care about.


Xyene (approved):
:100: &times; :+1:

PR #782, opened 2020-12-23
Guard against large integers in checker and interactor handling

quantum5:


Xyene (approved):

Xyene:
We can ignore the build failures, but it currently fails mypy too

PR #783, opened 2020-12-24
docker: temporarily work around non-POSIX rustup script

Xyene:


quantum5 (approved):
I don't think we need it in all three images, but it can't hurt.

PR #789, opened 2020-12-25
Unify all pt_debugger_* classes and switch to PTRACE_GETREGSET

quantum5:
This removes the need for architecture detection to use ptbox: in fact, processes could switch architectures half way if they wanted to.

Furthermore, support for running a seccomp build with seccomp disabled is added.


dmoj/cptbox/_cptbox.pyx
```diff
@@ -183,16 +190,13 @@ def bsd_get_proc_fdno(pid_t pid, int fd):
...
-    cdef int _debugger_type
 
-    property type:
-        def __get__(self):
-            return self._debugger_type
+    def __cinit__(self):
+        self.thisptr = new pt_debugger()
 
-    property getpid_syscall:
+    property skip_syscall_id:
```
Xyene, line 197: Maybe `noop_syscall_id` is more truthful, since the syscall itself isn't skipped.

dmoj/cptbox/_cptbox.pyx
```diff
@@ -417,25 +423,25 @@ cdef class Process:
...
-        for i in xrange(MAX_SYSCALL_NUMBER):
-            # We have to force exit syscalls to trap, so that we can be notified
-            # that the process spawned successfully when using `seccomp`. Otherwise,
-            # a simple assembly program could terminate without ever trapping.
-            if not self.debugger.is_exit(i):
-                config.syscall_whitelist[i] = self._syscall_whitelist[i]
-            else:
-                config.syscall_whitelist[i] = False
+        if self.use_seccomp:
+            config.avoid_seccomp = False
```
Xyene, line 430: It would be nice to have the `config` field also called `use_seccomp` for symmetry (and require less mental remapping)

dmoj/cptbox/_cptbox.pyx
```diff
@@ -417,25 +423,25 @@ cdef class Process:
...
+            config.abi_for_seccomp, whitelist = self._get_seccomp_abi_whitelist()
+            config.seccomp_whitelist = <bint*>malloc(sizeof(bint) * MAX_SYSCALL_NUMBER)
+            for i in range(MAX_SYSCALL_NUMBER):
+                config.seccomp_whitelist[i] = i < len(whitelist) and whitelist[i]
```
Xyene, line 434: `i < len(whitelist)` can't ever be true, can it? I feel we probably want to assert `len(whitelist) == MAX_SYSCALL_NUMBER`, if anything

dmoj/cptbox/_cptbox.pyx
```diff
@@ -445,6 +451,13 @@ cdef class Process:
         self._exited = True
         return self._exitcode
 
+    property use_seccomp:
+        def __get__(self):
+            return self.process.seccomp()
```
Xyene, line 456: We really call this thing too many things... so far there's

* `seccomp()`
* `use_seccomp`
* `avoid_seccomp`
* `_seccomp`

dmoj/cptbox/ptdebug_arm64.cpp
```diff
@@ -3,109 +3,109 @@
...
-int pt_debugger_arm64::getpid_syscall() {
-    return 172;
-}
+#undef MAKE_ACCESSOR
 
-pt_debugger_arm64::pt_debugger_arm64() {
-#if PTBOX_SECCOMP
-    execve_id = 221;
-#else
-    // execve is actually 221, but...
```
Xyene, line 104: Maybe keep this comment line

dmoj/cptbox/ptdebug_arm64.cpp
```diff
@@ -3,109 +3,109 @@
...
-    return (int) peek_reg(ARM_x8);
-}
-
-long pt_debugger_arm64::result() {
-    return peek_reg(ARM_x0);
-}
-
-void pt_debugger_arm64::result(long value) {
-    poke_reg(ARM_x0, value);
+#define INVALID_ABI(source) fprintf(stderr, source ": Invalid ABI\n")
```
Xyene, line 46: I think this should be an assert(false) and dump core, it doesn't seem like a good idea to have `syscall()` return -1 with only a warning.

dmoj/cptbox/ptdebug_x64.cpp
```diff
@@ -2,96 +2,129 @@
...
-#define RSP 19
-#define SS 20
-#define FS_BASE 21
-#define GS_BASE 22
-#define DS 23
-#define ES 24
-#define FS 25
-#define GS 26
+#ifndef __X32_SYSCALL_BIT
+#   define __X32_SYSCALL_BIT 0x40000000UL
```
Xyene, line 15: Where is this not available?

dmoj/cptbox/ptdebug_x64.cpp
```diff
@@ -2,96 +2,129 @@
...
-#if PTBOX_FREEBSD
-    poke_reg(RAX, id);
-#else
-    poke_reg(ORIG_RAX, id);
-#endif
+void pt_debugger::pre_syscall() {
+    struct iovec iovec;
+    iovec.iov_base = &regs;
+    iovec.iov_len = sizeof regs;
+    if (ptrace(PTRACE_GETREGSET, tid, NT_PRSTATUS, &iovec))
```
Xyene, line 32: Since we're touching this code anyway, let's add braces for peace of mind.

dmoj/cptbox/ptdebug_x64.cpp
```diff
@@ -2,96 +2,129 @@
...
+        perror("ptrace(PTRACE_GETREGSET)");
```
Xyene, line 33: I don't think `perror` is harsh enough for signalling that we're about to send up garbage.

dmoj/cptbox/ptdebug_x64.cpp
```diff
@@ -2,96 +2,129 @@
...
-    return 20;
-#else
-    return 39;
-#endif
-}
+#undef MAKE_ACCESSOR
 
-pt_debugger_x64::pt_debugger_x64() {
-    execve_id = 59;
+int pt_debugger::first_execve_syscall() {
```
Xyene, line 127: `execve_syscall_id` to match the naming of `(skip|noop)_syscall_id` from before?

dmoj/cptbox/tracer.py
```diff
@@ -30,29 +29,28 @@
...
-    (ARCH_ARM, ARCH_ARM): DEBUGGER_ARM,
-    (ARCH_A64, ARCH_ARM): DEBUGGER_ARM,
-    (ARCH_A64, ARCH_A64): DEBUGGER_ARM64,
+_SYSCALL_INDICIES: List[Optional[int]] = [None] * 5
+
+_SYSCALL_INDICIES[PTBOX_ABI_X86] = 0
+_SYSCALL_INDICIES[PTBOX_ABI_X64] = 1
+_SYSCALL_INDICIES[PTBOX_ABI_X32] = 2
+_SYSCALL_INDICIES[PTBOX_ABI_ARM] = 3
+_SYSCALL_INDICIES[PTBOX_ABI_ARM64] = 5
```
Xyene, line 38: Why skip 4?

quantum5: It's FreeBSD which is not actually supported as an ABI for now.

dmoj/cptbox/ptdebug_arm64.cpp
quantum5: Do we need to keep that line? It's obvious in the `return` that `execve` is 221.

dmoj/cptbox/_cptbox.pyx
```diff
@@ -202,7 +211,7 @@ cdef class Debugger:
             # When using seccomp, -1 as syscall means "skip"; when we are not,
             # we swap with a harmless syscall without side-effects (getpid).
             if not PTBOX_SECCOMP and value == -1:
```
Xyene, line 213: Compile-time `PTBOX_SECCOMP`, not `use_seccomp`?

quantum5: Fixed.

dmoj/cptbox/ptdebug_x64.cpp
```diff
@@ -2,96 +2,136 @@
...
-#endif
+void pt_debugger::post_syscall() {
+    if (!regs_changed || abi_ == PTBOX_ABI_INVALID)
+        return;
+
+    struct iovec iovec;
+    iovec.iov_base = &regs;
+    iovec.iov_len = abi_ == PTBOX_ABI_X86 ? sizeof regs.x86 : sizeof regs.x64;
+    if (ptrace(PTRACE_SETREGSET, tid, NT_PRSTATUS, &iovec)) {
+        perror("ptrace(PTRACE_SETREGSET)");
```
Xyene, line 52: We could've failed to change a syscall to a no-op here, is this ok?

dmoj/cptbox/ptproc.cpp
```diff
@@ -103,6 +95,10 @@ int pt_process::protection_fault(int syscall) {
     return PTBOX_EXIT_PROTECTION;
 }
 
+void pt_process::use_seccomp(bool enabled) {
+    _use_seccomp = PTBOX_SECCOMP && enabled;
```
Xyene, line 99: This function should fail if the process has already been started, otherwise `PTRACE_O_TRACESECCOMP` "sticks" on the process, but our handling could change.

dmoj/cptbox/tracer.py
```diff
@@ -87,51 +97,34 @@ def readstr(self, address, max_size=4096):
...
-        fsize=0,
-        address_grace=4096,
-        data_grace=0,
-        personality=0,
-        cwd='',
-        fds=None,
-        wall_time=None,
+            self,
+            args,
+            avoid_seccomp=False,
```
Xyene, line 107:
```suggestion
            use_seccomp=True,
```

dmoj/cptbox/tracer.py
```diff
@@ -87,51 +97,34 @@ def readstr(self, address, max_size=4096):
...
+            data_grace=0,
+            personality=0,
+            cwd='',
+            fds=None,
+            wall_time=None,
     ):
-        self._debugger_type = debugger
-        self._syscall_index = index = _SYSCALL_INDICIES[debugger]
         self._executable = executable or find_exe_in_path(args[0])
+        self.use_seccomp = security is not None and not avoid_seccomp
```
Xyene, line 126:
```suggestion
        self.use_seccomp = use_seccomp and security is not None
```

dmoj/cptbox/tracer.py
```diff
@@ -192,6 +181,23 @@ def __init__(
         if self._spawn_error:
             raise self._spawn_error
 
+    def _get_seccomp_abi_whitelist(self):
```
Xyene, line 184: `_get_seccomp_abi_and_whitelist` :)

Xyene (commented):
Otherwise LGTM

dmoj/cptbox/ptdebug_x64.cpp
quantum5: Not quite, but it's no worse than before, when updating every register could fail. We need to think of a way to report the errors, but that's probably for later.

dmoj/cptbox/tracer.py
quantum5: Disagree, because use_seccomp=True can still not use seccomp on non-seccomp builds.

dmoj/cptbox/tracer.py
quantum5: Also disagree because of semantic issues with `use_seccomp`.

PR #796, opened 2020-12-27
Re-add FreeBSD support

quantum5:
Properly support FreeBSD once again after `seccomp` broke it.


quantum5:
Tested on actual FreeBSD box and all sandbox features appear to work.

Xyene (approved):

PR #808, opened 2020-12-29
Build C extensions in parallel

quantum5:
Platform | Single time | Parallel time | Speed up | Saved
-- | -- | -- | -- | --
Ryzen 9 3900X | 3.743 | 2.557 | 1.463824795 | 1.186
Raspberry Pi 3 (armhf) | 73.19 | 42.921 | 1.70522588 | 30.269
Raspberry Pi 3 (arm64) | 83.418 | 47.355 | 1.761545771 | 36.063

While this offers a negligible improvement on fast machines (saves around 1 second), it can mean a lot on slower multicore machines, such as the Raspberry Pi 3. The improvements would be more impressive if only `_cptbox` was the only extension being built.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,10 @@ import io
 import os
 import sys
 import traceback
+from distutils.ccompiler import CCompiler
 from distutils.errors import DistutilsPlatformError
+from multiprocessing import cpu_count
+from multiprocessing.pool import ThreadPool
 
 from setuptools import Extension, find_packages, setup
 from setuptools.command.build_ext import build_ext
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -22,6 +25,13 @@
 
 # Allow manually disabling seccomp on old kernels. WSL doesn't have seccomp.
 has_seccomp = sys.platform.startswith('linux') and not is_wsl and os.environ.get('DMOJ_USE_SECCOMP') != 'no'
+try:
+    parallel = int(os.environ['DMOJ_PARALLEL'])
+except (KeyError, ValueError):
+    parallel = cpu_count()
```
kiritofeng, line 31: This can fail with `NotImplementedError`, should we provide a default value of 1 if it does so?

quantum5: Good catch.

kiritofeng (approved):
LGTM

PR #811, opened 2020-12-29
Factor out shared pre-/post-syscall code

quantum5:


dmoj/cptbox/ptdebug.cpp
```diff
@@ -49,6 +53,50 @@ void pt_debugger::settid(pid_t tid) {
...
+#else
+        abi_ = abi_from_reg_size(iovec.iov_len);
+#endif
+        regs_changed = false;
+        return true;
+    }
+}
+
+bool pt_debugger::post_syscall() {
+    if (!regs_changed || abi_ == PTBOX_ABI_INVALID)
```
Xyene, line 82:
```suggestion
    assert(abi_ != PTBOX_ABI_INVALID);
    if (!regs_changed)
```

+ a comment saying that this should already be handled?

Xyene (approved):

PR #814, opened 2020-12-30
Ignore pre-syscall ESRCH which may happen on exit

quantum5:


WallE256 (approved):

Xyene (approved):
LGTM, though I'd like a comment to the effect of what you explained in chat

PR #818, opened 2020-12-31
Add tests for #760

kiritofeng:


WallE256 (approved):

testsuite/helloworld/tests/go_low_mem/test.yml
```diff
@@ -0,0 +1,5 @@
+language: GO
+time: 2
+memory: 6144
```
WallE256, line 3: It fails on one of the tests using around 7MB https://github.com/DMOJ/judge-server/pull/818/checks?check_run_id=1631528366#step:8:560

PR #841, opened 2021-03-06
Make bridged interactors unbuffered by default

kiritofeng:


testsuite/bridged_interactor_noflush/init.yml
```diff
@@ -0,0 +1,9 @@
+unbuffered: True
```
Xyene, line 1: BRO

Xyene (changes requested):

testsuite/bridged_interactor_noflush/init.yml
kiritofeng: if it make you feel worse, there's one that says `unbuffered: yes`

Xyene: It doesn't, because `yes` is actually part of the YAML spec.

kiritofeng: big TIL moment

Xyene (approved):

PR #842, opened 2021-03-12
Allow for `str` batch names

kiritofeng:
Consider the following:
```
dummy.1.{in,out}
dummy.2.{in,out}
1.{in,out}
```
(this setup is common with COCI)

currently, this will cause an IE when the sort tries to compare `1` with `'dummy'`, and requires manual intervention (e.g. renaming the files to `0.1.{in,out}`).


dmoj/problem.py
```diff
@@ -104,7 +104,7 @@ class _TestCase:
                 setattr(groups[batch][case], filetype, testcase_file)
 
         test_cases = []
-        for batch_or_case_id in sorted(groups.keys()):
+        for batch_or_case_id in sorted(groups.keys(), key=lambda id : (isinstance(id, int), id)):
```
Xyene, line 107: I'm surprised this passed linting...

```suggestion
        for batch_or_case_id in sorted(groups.keys(), key=lambda id: (isinstance(id, int), id)):
```

otherwise LGTM

Xyene (approved):

PR #843, opened 2021-03-29
Allow for checkers to call `int()`

kiritofeng:
The current behaviour requires an argument to be passed to `int(...)`, however, this breaks `defaultdict(int)`.

### DIFF
--- a/dmoj/tests/test_int_patch.py
+++ b/dmoj/tests/test_int_patch.py
@@ -39,6 +39,7 @@ class IntrospectionTest(unittest.TestCase):
         self.assertEqual(type(1), int)
 
     def test_identity(self):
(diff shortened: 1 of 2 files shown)


Ninjaclasher (approved):

WallE256 (approved):

Xyene (approved):

Xyene:
LGTM, but could you also add a test for this in https://github.com/DMOJ/judge-server/blob/master/dmoj/tests/test_int_patch.py?

quantum5 (changes requested):
Looks good, but will not merge unless there is a test.

quantum5 (approved):

PR #844, opened 2021-03-30
Add checker for problems that reward points based on lines matched

Carson-Tang:
There are problems such as https://dmoj.ca/problem/coci09c4p4 that reward a % of points for the test case if the output is partially correct.

### DIFF
--- a/dmoj/checkers/linematches.py
+++ b/dmoj/checkers/linematches.py
@@ -1,0 +1,31 @@
+from re import split as resplit
+from typing import List, Union
+
+from dmoj.error import InternalError
+from dmoj.result import CheckerResult
+from dmoj.utils.unicode import utf8bytes
+
+
+def check(
+    process_output: bytes, judge_output: bytes, point_value: float = 1,
+    point_distribution: List[int] = [1], filler_lines_required: bool = True, **kwargs
+) -> Union[CheckerResult, bool]:
+    judge_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(judge_output))))
+
+    if len(judge_lines) != len(point_distribution):
+        raise InternalError('point distribution length must equal to judge output length')
+
+    if sum(point_distribution) == 0:
+        raise InternalError('sum of point distribution must be positive')
+
+    process_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(process_output))))
+
+    if filler_lines_required and len(process_lines) != len(judge_lines):
+        return False
+
+    points = 0
+    for process_line, judge_line, line_points in zip(process_lines, judge_lines, point_distribution):
+        if process_line == judge_line:
+            points += line_points
+
+    return CheckerResult(points > 0, point_value * (points / sum(point_distribution)))
--- a/dmoj/tests/test_checker.py
+++ b/dmoj/tests/test_checker.py
@@ -1,5 +1,7 @@
 import unittest
 
+from dmoj.result import CheckerResult
+
 
 class CheckerTest(unittest.TestCase):
     def assert_pass(self, check, str1, str2, expect=True):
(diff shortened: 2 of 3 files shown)


Xyene:
ok to test

quantum5:
add to whitelist

dmoj/checkers/linematches.py
```diff
@@ -0,0 +1,24 @@
...
+    judge_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(judge_output))))
+
+    if filler_lines_required and len(process_lines) != len(judge_lines):
+        return False
+
+    points = 0
+
+    for i, (process_line, judge_line) in enumerate(zip(process_lines, judge_lines)):
+        if process_line == judge_line:
+            points += point_distribution[i] if i < len(point_distribution) else 0
```
Xyene, line 22: Semantically speaking, shouldn't this be WA if `len(point_distributions) != len(process_lines)`? Can `len(judge_lines) != len(point_distribution)`? What would that mean?

Carson-Tang: It's possible for `len(point_distributions) != len(process_lines)` to still be AC (get partial points) as there are some problems which don't require the user to output the complete number of lines, 
see https://dmoj.ca/problem/coci11c2p1 - "even if the second line of output is not present".

Ideally, `len(point_distributions)` and `len(judge_lines)` should always be equal, which would be the responsibility of the user configuring the problem yml. Should it just assume they are equal and IE if the problem creator fails to provide a valid point distribution?

There's also the issue where if the `point_distribution` passed is `[]` or `[0]` in which case the checker would just die from division by zero.

Xyene: Let's fail if the distribution and number of judge lines are not matched, or if we'd end up dividing by zero. A helpful error message would be, well, more helpful to a problemsetter than a random Python exception.

Carson-Tang: Sounds good, changes have been made.

dmoj/checkers/linematches.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def check(
+    process_output: bytes, judge_output: bytes, point_value: float = 1,
+    point_distribution: List[int] = [1], filler_lines_required: bool = True, **kwargs
+) -> Union[CheckerResult, bool]:
+    process_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(process_output))))
+    judge_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(judge_output))))
+
+    if len(judge_lines) != len(point_distribution):
```
Xyene, line 16: Nit: we can do this before wasting time processing the process output.

Xyene: Micronit: usually our IEs start with lowercase letters (though there are a few exceptions).

Carson-Tang: Resolved

dmoj/checkers/linematches.py
```diff
@@ -0,0 +1,32 @@
...
+        raise InternalError('sum of point distribution must be positive')
+
+    process_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(process_output))))
+
+    if filler_lines_required and len(process_lines) != len(judge_lines):
+        return False
+
+    points = 0
+
+    for i, (process_line, judge_line) in enumerate(zip(process_lines, judge_lines)):
```
quantum5, line 28: Why not just `zip(process_lines, judge_lines, point_distribution)`?

Carson-Tang: True... thanks for the suggestion, it's been updated.

dmoj/checkers/linematches.py
```diff
@@ -0,0 +1,32 @@
...
+        raise InternalError('sum of point distribution must be positive')
+
+    process_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(process_output))))
+
+    if filler_lines_required and len(process_lines) != len(judge_lines):
+        return False
+
+    points = 0
+
+    for process_line, judge_line, point in zip(process_lines, judge_lines, point_distribution):
```
kiritofeng, line 28:
```suggestion
    for process_line, judge_line, line_points in zip(process_lines, judge_lines, point_distribution):
```

dmoj/checkers/linematches.py
```diff
@@ -0,0 +1,32 @@
...
+    if sum(point_distribution) == 0:
+        raise InternalError('sum of point distribution must be positive')
+
+    process_lines = list(filter(None, resplit(b'[\r\n]', utf8bytes(process_output))))
+
+    if filler_lines_required and len(process_lines) != len(judge_lines):
+        return False
+
+    points = 0
+
```
Xyene, line 27:
```suggestion
```

Xyene (commented):
Apart from the nit from @kiritofeng, this LGTM.

dmoj/tests/test_checker.py
```diff
@@ -63,3 +65,20 @@ def test_sorted(self):
         assert check(b'1 2\n3', b'3\n1 2')
         assert not check(b'1 2\n3', b'3\n2 1')
         assert check(b'1 2\n3', b'3\n2 1', split_on='whitespace')
+
+    def test_linematches(self):
+        from dmoj.checkers.linematches import check
+
+        def validate(res, expected_points, passed):
```
quantum5, line 72: This is not the right style of testing. Checks should be done with `self.assert_*`. This can be fixed later given how broken this file is.

quantum5 (approved):

PR #847, opened 2021-04-13
Give zig more time to compile

quantum5:


kiritofeng (approved):
I think we usually put that before the `test_program` but otherwise lgtm

PR #855, opened 2021-05-03
Add python 3.9 classifier to setup.py

quantum5:
We released 3.0.x after adding 3.9 support and forgot to tag it properly.


kiritofeng (approved):
LGTM

PR #856, opened 2021-05-07
Add executor for Pyston

UserBlackBox:
Added executor for [Pyston](https://github.com/pyston/pyston) interpreter. I tested it with `dmoj-cli`

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ The judge can also grade in the languages listed below. These languages are less
 * V8 JavaScript
 * Brain\*\*\*\*
 * Zig
+* Pyston
 
 ## Installation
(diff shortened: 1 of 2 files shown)


fluix-dev:
I think this would put more burden on problem setters to put specific limits for Python, PyPy, and now Pyston for little gain. This is also a rather new interpreter if I understand correctly, right?

fluix-dev:
Could you compare it to other languages? Paste the source and results here and if it's similar enough to other languages then yeah, I think I can get behind this.

UserBlackBox:
Here is a test for the problem https://dmoj.ca/problem/dpb
image
https://dmoj.ca/src/1682664

https://dmoj.ca/src/1682668

I guess pyston isn't optimized for competitive programming type of tasks which makes sense considering its made for web servers. No point in adding another PyPy.

PR #861, opened 2021-05-21
Raise IE when no test cases are found; fixes #860

### LINKED ISSUE #860: Raise IE when no testcases are found

kiritofeng:


dmoj/tests/test_problem.py
```diff
@@ -96,5 +96,20 @@ def test_bad_archive(self):
             with self.assertRaisesRegex(InvalidInitException, 'bad archive:'):
                 Problem('test', 2, 16384, {})
 
+    def test_no_testcases(self):
+        class MockProblem(Problem):
```
quantum5, line 100: Why do we need this class?

quantum5 (commented):
Also, what happens when you have a batch, but it has no cases?

dmoj/tests/test_problem.py
kiritofeng: It tries to list the files in the archive it will try the regex), and if you don't fake the archive's existence it errors out by saying the archive does not exist

quantum5 (approved):

quantum5:
@kiritofeng should probably rebase this so that FreeBSD runs.

Issue #871, opened 2021-08-28
Landlock support


Xyene:
Linux 5.13 introduced a new security subsystem, [Landlock](https://www.kernel.org/doc/html/latest/userspace-api/landlock.html), that we can make use of in cptbox for filesystem sandboxing.

Currently, cptbox in the standard configuration makes use of seccomp to whitelist syscalls, and trap into ptrace for syscalls that need either sanitization or unconditional failure with an error code (e.g., `socket`). Landlock would allow us to drop ptrace entirely, and let the kernel do more of the heavy lifting. A ptrace-less implementation would also remove the need for multiprocessed programs to take a performance hit.

A non-exhaustive checklist of things that need to happen:

* [ ] Filesystem filters need to be rewritten to not use regex. Landlock provides "allow file read" and "allow (recursive) directory read" primitives; these are sufficient for our purposes (and can be converted to regex for the ptrace + seccomp backend). A first step would be to get this working with the ptrace + seccomp backend, before starting work on the Landlock bits.

`BASE_FILESYSTEM` is currently defined as

```python
BASE_FILESYSTEM = [
    '/dev/(?:null|tty|zero|u?random)$',
    '/usr/(?!home)',
    '/lib(?:32|64)?/',
    '/opt/',
    '/etc$',
    '/etc/(?:localtime|timezone)$',
    '/usr$',
    '/tmp$',
    '/$',
]
```

A Landlock-friendly version could look something like

```python
BASE_FILESYSTEM = Root((
    Dir("usr", (
        Self(),
        All(except=(Dir("home"),)),
    )),
    Dir("tmp", (Self(),)),
    Dir("etc", (
        Self(),
        File("localtime"),
        File("timezone"),
    )),
    Dir("dev", (
        File("null"),
        File("tty"),
        File("zero"),
        File("urandom"),
        File("random"),
    )),
    Dir("lib", (All(),))
    Dir("lib32", (All(),))
    Dir("lib64", (All(),))
    Dir("opt", (All(),))
))
```

These objects should support composition:

```python
a = Root((
    Dir("etc", (
        File("nsswitch.conf")))))
b = Root((
    Dir("etc", (
        File("passwd")))))
a + b
```

should be equivalent to

```python
Root((
    Dir("etc", (
        File("nsswitch.conf"),
        File("passwd")))))
```

[...]

PR #873, opened 2021-08-30
change regex to filesystem policy matcher, first step of #871

Riolku:
This is the first step of migrating to Landlock support. A very simplistic implementation is given here.

### DIFF
--- a/.freebsd.test.py
+++ b/.freebsd.test.py
@@ -31,13 +31,18 @@ OVERRIDES = {}
 
 def main():
     judgeenv.env['runtime'] = {}
-    judgeenv.env['extra_fs'] = {'PERL': ['/dev/dtrace/helper$'], 'RUBY2': ['/dev/dtrace/helper$']}
+    judgeenv.env['extra_fs'] = {
+        'PERL': [{'exact_file': '/dev/dtrace/helper'}],
+        'RUBY2': [{'exact_file': '/dev/dtrace/helper'}],
+    }
 
     logging.basicConfig(level=logging.INFO)
 
     print('Using extra allowed filesystems:')
     for lang, fs in judgeenv.env['extra_fs'].iteritems():
-        print('%-6s: %s' % (lang, '|'.join(fs)))
+        for rules in fs:
+            for access_type, file in rules.iteritems():
+                print('%-6s: %s: %s' % (lang, access_type, file))
     print()
 
     print('Testing executors...')
--- a/dmoj/cptbox/filesystem_policies.py
+++ b/dmoj/cptbox/filesystem_policies.py
@@ -1,0 +1,117 @@
+import os
+from enum import Enum
+from typing import List, Union
+
+
+class AccessMode(Enum):
+    NONE = 0
+    EXACT = 1
+    RECURSIVE = 2
+
+    # flake8 doesn't recognize `AccessMode` in type annotations here
+    @classmethod
+    def more_permissive(cls, mode1, mode2):
+        return cls(max(mode1.value, mode2.value))
+
+
+class Dir:
+    def __init__(self):
+        self.access_mode = AccessMode.NONE
+        self.subpath_map = {}
+
+
+class File:
+    pass
+
+
+class ExactFile:
+    def __init__(self, path: str):
+        self.path = path
+
+
+class ExactDir:
+    access_mode = AccessMode.EXACT
+
+    def __init__(self, path: str):
+        self.path = path
+
+
+class RecursiveDir:
+    access_mode = AccessMode.RECURSIVE
+
+    def __init__(self, path: str):
+        self.path = path
+
+
+FilesystemAccessRule = Union[ExactFile, ExactDir, RecursiveDir]
+
+
+class FilesystemPolicy:
+    def __init__(self, rules: List[FilesystemAccessRule]):
+        self.root = Dir()
+        for rule in rules:
+            self._add_rule(rule)
+
+    def _add_rule(self, rule: FilesystemAccessRule) -> None:
+        self._assert_rule_type(rule)
+        if rule.path == '/':
+            return self._finalize_root_rule(rule)
+
+        path = rule.path
+        assert os.path.abspath(path) == path, 'FilesystemAccessRule must specify a normalized, absolute path to rule'
+        *directory_path, final_component = path.split('/')[1:]
+
+        node = self.root
+        for component in directory_path:
+            new_node = node.subpath_map.setdefault(component, Dir())
+            assert isinstance(new_node, Dir), 'Cannot add rule: refusing to descend into non-directory'
+            node = new_node
+
+        self._finalize_rule(node, final_component, rule)
+
+    def _assert_rule_type(self, rule: FilesystemAccessRule) -> None:
+        if os.path.exists(rule.path):
+            is_dir = os.path.isdir(rule.path)
+            if isinstance(rule, ExactFile):
+                assert not is_dir, f"Can't apply file rule to directory {rule.path}"
+            else:
+                assert is_dir, f"Can't apply directory rule to non-directory {rule.path}"
+
+    def _finalize_root_rule(self, rule: FilesystemAccessRule) -> None:
+        assert not isinstance(rule, ExactFile), 'Root is not a file'
+        self._finalize_directory_rule(self.root, rule)
+
+    def _finalize_rule(self, node: Dir, final_component: str, rule: FilesystemAccessRule) -> None:
+        assert final_component != '', 'Must not have trailing slashes in rule path'
+        if isinstance(rule, ExactFile):
+            new_node = node.subpath_map.setdefault(final_component, File())
+            assert isinstance(new_node, File), "Can't add ExactFile: Dir rule exists"
+        else:
+            new_node = node.subpath_map.setdefault(final_component, Dir())
+            assert isinstance(new_node, Dir), "Can't add rule: File rule exists"
+            self._finalize_directory_rule(new_node, rule)
+
+    def _finalize_directory_rule(self, node: Dir, rule: Union[ExactDir, RecursiveDir]) -> None:
+        node.access_mode = AccessMode.more_permissive(
+            node.access_mode, rule.access_mode
+        )  # Allow the more permissive rule
+
+    # `path` should be a normalized path
+    def check(self, path: str) -> bool:
+        assert os.path.abspath(path) == path, 'Must pass a normalized, absolute path to check'
+
+        node = self.root
+        for component in path.split('/')[1:]:
+            if isinstance(node, File):
+                return False
+            elif node.access_mode == AccessMode.RECURSIVE:
+                return True
+            else:
+                node = node.subpath_map.get(component)
+                if node is None:
+                    return False
+
+        return self._check_final_node(node)
+
+    def _check_final_node(self, node: Union[Dir, File]) -> bool:
+        return isinstance(node, File) or node.access_mode != AccessMode.NONE
--- a/dmoj/cptbox/isolate.py
+++ b/dmoj/cptbox/isolate.py
@@ -1,9 +1,9 @@
 import logging
 import os
-import re
 import sys
 
 from dmoj.cptbox._cptbox import AT_FDCWD, bsd_get_proc_cwd, bsd_get_proc_fdno
+from dmoj.cptbox.filesystem_policies import FilesystemPolicy
 from dmoj.cptbox.handlers import ACCESS_EACCES, ACCESS_ENAMETOOLONG, ACCESS_ENOENT, ACCESS_EPERM, ALLOW
 from dmoj.cptbox.syscalls import *
 from dmoj.cptbox.tracer import MaxLengthExceeded
@@ -183,12 +183,7 @@ class IsolateTracer(dict):
             )
 
     def _compile_fs_jail(self, fs):
-        if fs:
-            fs_re = '|'.join(fs)
-        else:
-            fs_re = '(?!)'  # Disallow accessing everything by default.
-
-        return re.compile(fs_re)
+        return FilesystemPolicy(fs or [])
 
     def is_write_flags(self, open_flags):
         for flag in open_write_flags:
@@ -257,9 +252,7 @@ class IsolateTracer(dict):
 
         is_write = is_open and self.is_write_flags(getattr(debugger, 'uarg%d' % flag_reg))
         fs_jail = self.write_fs_jail if is_write else self.read_fs_jail
-        if fs_jail.match(file) is None:
-            return file, False
-        return file, True
+        return file, fs_jail.check(file)
 
     def get_full_path(self, debugger, file, dirfd=AT_FDCWD):
         dirfd = (dirfd & 0x7FFFFFFF) - (dirfd & 0x80000000)
--- a/dmoj/executors/COFFEE.py
+++ b/dmoj/executors/COFFEE.py
@@ -1,5 +1,6 @@
 import os
 
+from dmoj.cptbox.filesystem_policies import ExactFile
 from dmoj.executors.script_executor import ScriptExecutor
 
 
@@ -38,7 +39,7 @@ process.stdin.on 'readable', () ->
         return [self.get_command(), self.runtime_dict['coffee'], self._code]
 
     def get_fs(self):
-        return super().get_fs() + [self.runtime_dict['coffee'], self._code]
+        return super().get_fs() + [ExactFile(self.runtime_dict['coffee']), ExactFile(self._code)]
 
     @classmethod
     def get_versionable_commands(cls):
--- a/dmoj/executors/DART.py
+++ b/dmoj/executors/DART.py
@@ -16,7 +16,6 @@ void main() {
     address_grace = 128 * 1024
 
     syscalls = ['epoll_create', 'epoll_ctl', 'epoll_wait', 'timerfd_settime', 'memfd_create', 'ftruncate']
-    fs = ['.*/vm-service$']
 
     def get_compile_args(self):
         return [self.get_command(), '--snapshot=%s' % self.get_compiled_file(), self._code]
--- a/dmoj/executors/SWIFT.py
+++ b/dmoj/executors/SWIFT.py
@@ -5,7 +5,6 @@ class Executor(CompiledExecutor):
     ext = 'swift'
     name = 'SWIFT'
     command = 'swiftc'
-    fs = ['/lib']
     test_program = 'print(readLine()!)'
 
     def get_compile_args(self):
--- a/dmoj/executors/mixins.py
+++ b/dmoj/executors/mixins.py
@@ -6,6 +6,7 @@ import sys
 from typing import Any, List, Tuple, Union
 
 from dmoj.cptbox import IsolateTracer, TracedPopen, syscalls
+from dmoj.cptbox.filesystem_policies import ExactDir, ExactFile, FilesystemAccessRule, RecursiveDir
 from dmoj.cptbox.handlers import ALLOW
 from dmoj.error import InternalError
 from dmoj.judgeenv import env
@@ -13,37 +14,58 @@ from dmoj.utils import setbufsize_path
 from dmoj.utils.unicode import utf8bytes
 
 BASE_FILESYSTEM = [
-    '/dev/(?:null|tty|zero|u?random)$',
-    '/usr/(?!home)',
-    '/lib(?:32|64)?/',
-    '/opt/',
-    '/etc$',
-    '/etc/(?:localtime|timezone)$',
-    '/usr$',
-    '/tmp$',
-    '/$',
+    ExactFile('/dev/null'),
+    ExactFile('/dev/tty'),
+    ExactFile('/dev/zero'),
+    ExactFile('/dev/urandom'),
+    ExactFile('/dev/random'),
+    *[RecursiveDir(f'/usr/{d}') for d in os.listdir('/usr') if d != 'home' and os.path.isdir(f'/usr/{d}')],
+    RecursiveDir('/lib'),
+    RecursiveDir('/lib32'),
+    RecursiveDir('/lib64'),
+    RecursiveDir('/opt'),
+    ExactDir('/etc'),
+    ExactFile('/etc/localtime'),
+    ExactFile('/etc/timezone'),
+    ExactDir('/usr'),
+    ExactDir('/tmp'),
+    ExactDir('/'),
 ]
-BASE_WRITE_FILESYSTEM = ['/dev/stdout$', '/dev/stderr$', '/dev/null$']
+
+BASE_WRITE_FILESYSTEM = [ExactFile('/dev/stdout'), ExactFile('/dev/stderr'), ExactFile('/dev/null')]
+
 
 if 'freebsd' in sys.platform:
-    BASE_FILESYSTEM += [r'/etc/s?pwd\.db$', '/dev/hv_tsc$']
+    BASE_FILESYSTEM += [ExactFile('/etc/spwd.db'), ExactFile('/etc/pwd.db'), ExactFile('/dev/hv_tsc')]
+
 else:
-    BASE_FILESYSTEM += ['/sys/devices/system/cpu(?:$|/online)', '/etc/selinux/config$']
+    BASE_FILESYSTEM += [
+        ExactDir('/sys/devices/system/cpu'),
+        ExactFile('/sys/devices/system/cpu/online'),
+        ExactFile('/etc/selinux/config'),
+    ]
 
 if sys.platform.startswith('freebsd'):
-    BASE_FILESYSTEM += [r'/etc/libmap\.conf$', r'/var/run/ld-elf\.so\.hints$']
+    BASE_FILESYSTEM += [ExactFile('/etc/libmap.conf'), ExactFile('/var/run/ld-elf.so.hints')]
 else:
     # Linux and kFreeBSD mounts linux-style procfs.
     BASE_FILESYSTEM += [
-        '/proc$',
-        '/proc/self/(?:maps|exe|auxv)$',
-        '/proc/self$',
-        '/proc/(?:meminfo|stat|cpuinfo|filesystems|xen|uptime)$',
-        '/proc/sys/vm/overcommit_memory$',
+        ExactDir('/proc'),
+        ExactDir('/proc/self'),
+        ExactFile('/proc/self/maps'),
+        ExactFile('/proc/self/exe'),
+        ExactFile('/proc/self/auxv'),
+        ExactFile('/proc/meminfo'),
+        ExactFile('/proc/stat'),
+        ExactFile('/proc/cpuinfo'),
+        ExactFile('/proc/filesystems'),
+        ExactDir('/proc/xen'),
+        ExactFile('/proc/uptime'),
+        ExactFile('/proc/sys/vm/overcommit_memory'),
     ]
 
     # Linux-style ld.
-    BASE_FILESYSTEM += [r'/etc/ld\.so\.(?:nohwcap|preload|cache)$']
+    BASE_FILESYSTEM += [ExactFile('/etc/ld.so.nohwcap'), ExactFile('/etc/ld.so.preload'), ExactFile('/etc/ld.so.cache')]
 
 UTF8_LOCALE = 'C.UTF-8'
 
@@ -56,8 +78,8 @@ class PlatformExecutorMixin(metaclass=abc.ABCMeta):
     data_grace = 0
     fsize = 0
     personality = 0x0040000  # ADDR_NO_RANDOMIZE
-    fs: List[str] = []
-    write_fs: List[str] = []
+    fs: List[FilesystemAccessRule] = []
+    write_fs: List[FilesystemAccessRule] = []
     syscalls: List[Union[str, Tuple[str, Any]]] = []
 
     def _add_syscalls(self, sec):
@@ -74,9 +96,20 @@ class PlatformExecutorMixin(metaclass=abc.ABCMeta):
         return self._add_syscalls(sec)
 
     def get_fs(self):
+        return BASE_FILESYSTEM + self.fs + self._load_extra_fs() + [RecursiveDir(self._dir)]
+
+    def _load_extra_fs(self):
         name = self.get_executor_name()
-        fs = BASE_FILESYSTEM + self.fs + env.get('extra_fs', {}).get(name, []) + [re.escape(self._dir)]
-        return fs
+        extra_fs_config = env.get('extra_fs', {}).get(name, [])
+        extra_fs = []
+        constructors = dict(exact_file=ExactFile, exact_dir=ExactDir, recursive_dir=RecursiveDir)
+        for rules in extra_fs_config:
+            for type, path in rules.iteritems():
+                constructor = constructors.get(type)
+                assert constructor, f"Can't load rule for extra path with rule type {type}"
+                extra_fs.append(constructor(path))
+
+        return extra_fs
 
     def get_write_fs(self):
         return BASE_WRITE_FILESYSTEM + self.write_fs
--- a/dmoj/executors/mono_executor.py
+++ b/dmoj/executors/mono_executor.py
@@ -3,6 +3,7 @@ import re
 from collections import deque
 
 from dmoj.cptbox import TracedPopen
+from dmoj.cptbox.filesystem_policies import RecursiveDir
 from dmoj.cptbox.handlers import ACCESS_EAGAIN
 from dmoj.executors.compiled_executor import CompiledExecutor
 from dmoj.result import Result
@@ -32,7 +33,7 @@ class MonoExecutor(CompiledExecutor):
     # get flagged as MLE.
     data_grace = 65536
     cptbox_popen_class = MonoTracedPopen
-    fs = ['/etc/mono/']
+    fs = [RecursiveDir('/etc/mono')]
     # Mono sometimes forks during its crashdump procedure, but continues even if
     # the call to fork fails.
     syscalls = [
--- a/dmoj/tests/test_filesystem_policy.py
+++ b/dmoj/tests/test_filesystem_policy.py
@@ -1,0 +1,62 @@
+import unittest
+
+from dmoj.cptbox.filesystem_policies import ExactDir, ExactFile, FilesystemPolicy, RecursiveDir
+
+
+class CheckerTest(unittest.TestCase):
+    def test_exact_dir(self):
+        self.fs = FilesystemPolicy([ExactDir('/etc')])
+
+        self.checkTrue('/etc')
+
+        self.checkFalse('/')
+        self.checkFalse('/etc/passwd')
+        self.checkFalse('/etc/shadow')
+
+    def test_recursive_dir(self):
+        self.fs = FilesystemPolicy([RecursiveDir('/usr')])
+
+        self.checkTrue('/usr')
+        self.checkTrue('/usr/lib')
+        self.checkTrue('/usr/lib/a/b/c/d/e')
+
+        self.checkFalse('/')
+        self.checkFalse('/etc')
+        self.checkFalse('/us')
+        self.checkFalse('/usr2')
+
+    def test_exact_file(self):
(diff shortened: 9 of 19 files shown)


Xyene:
ok to test

dmoj/executors/mixins.py
```diff
@@ -12,38 +12,76 @@
...
-    BASE_FILESYSTEM += [r'/etc/s?pwd\.db$', '/dev/hv_tsc$']
+    BASE_FILESYSTEM += FilesystemPolicy(
+        File("/etc/spwd.db"),
+        File("/etc/pwd.db"),
+        File("/dev/hv_tsc")
+    )
 else:
-    BASE_FILESYSTEM += ['/sys/devices/system/cpu(?:$|/online)', '/etc/selinux/config$']
+    BASE_FILESYSTEM += FilesystemPolicy(
+        File("/sys/devices/system/cpu"),
```
Xyene, line 53: `/sys/devices/system/cpu` is a directory, not a file.

dmoj/executors/filesystem_rules.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def check(self, path):
+        return any(item.check(path) for item in self.items)
+
+class Dir:
+    def __init__(self, path):
+        self.path = path
+
+    def check(self, path):
+        return path.startswith(self.path)
```
quantum5, line 16: Seems like a security issue? Allowing `/usr` would end up allowing access to `/usr2/stuff`.

dmoj/executors/mixins.py
```diff
@@ -12,38 +12,76 @@
...
     # Linux and kFreeBSD mounts linux-style procfs.
-    BASE_FILESYSTEM += [
-        '/proc$',
-        '/proc/self/(?:maps|exe|auxv)$',
-        '/proc/self$',
-        '/proc/(?:meminfo|stat|cpuinfo|filesystems|xen|uptime)$',
-        '/proc/sys/vm/overcommit_memory$',
-    ]
+    BASE_FILESYSTEM += FilesystemPolicy(
+        File("/proc"),
```
Xyene, line 66: Is a directory.

dmoj/executors/DART.py
```diff
@@ -16,7 +17,7 @@ class Executor(CompiledExecutor):
     address_grace = 128 * 1024
 
     syscalls = ['epoll_create', 'epoll_ctl', 'epoll_wait', 'timerfd_settime', 'memfd_create', 'ftruncate']
-    fs = ['.*/vm-service$']
+    fs = FilesystemPolicy()
```
Xyene, line 20: Don't need to specify a policy if it is the default.

dmoj/executors/mixins.py
```diff
@@ -56,8 +94,8 @@ class PlatformExecutorMixin(metaclass=abc.ABCMeta):
     data_grace = 0
     fsize = 0
     personality = 0x0040000  # ADDR_NO_RANDOMIZE
-    fs: List[str] = []
-    write_fs: List[str] = []
+    fs: List[str] = FilesystemPolicy()
+    write_fs: List[str] = FilesystemPolicy()
```
quantum5, line 98: Fix the type annotations.

dmoj/executors/filesystem_rules.py
Riolku: can I fix this with something as trivial as `path.startswith(self.path + "/")` or should we do something more intelligent

quantum5: Appending / will break stat on the directory itself, so you'll need something more intelligent.

Riolku: I guess the next dumb idea is `path == self.path or path.startswith(self.path + "/")`

Xyene: You probably want to compare `os.path.split(path)` with `os.path.split(self.path)`. Easier to think about per-token comparison.

Riolku: That sounds much more sane

Riolku: What did the regex system do?

dmoj/executors/filesystem_rules.py
```diff
@@ -0,0 +1,23 @@
+class FilesystemPolicy:
+    def __init__(self, *items):
+        self.items = list(items)
+
+    def __add__(self, other):
+        return FilesystemPolicy(*(self.items + other.items))
+
+    def check(self, path):
```
Xyene, line 8: I think `check` should also take an `is_file` parameter. If a rule for `Dir('/foo/bar)` exists but `/foo/bar` is actually a directory, it should be disallowed. We know `is_file` from the syscall that got triggered.

dmoj/executors/mixins.py
```diff
@@ -12,38 +12,76 @@
...
+    File("/dev/urandom"),
+    File("/dev/random"),
+    *[ Dir(f"/usr/{d}") for d in os.listdir("/usr") if d != "home" ],
+
+    Dir("/lib"),
+    Dir("/lib32"),
+    Dir("/lib64"),
+    Dir("/opt"),
+
+    File("/etc"),
```
Xyene, line 30: `/etc` is a directory.

dmoj/executors/filesystem_rules.py
Riolku: It specified the directory like `/usr$` and the full match `/usr/`, once each?

dmoj/executors/mixins.py
```diff
@@ -12,38 +12,76 @@
...
+    File("/etc/localtime"),
```
Xyene, line 31: But, either way this isn't actually consistent. Either `Dir('/etc')` means:

- "allow everything under `/etc`, recursively", which is scary, and it shouldn't, but in which case the `/etc/localtime` rules are duplicative, or
- it means "allow only /etc, recursively", in which case the `/etc/localtime` rules make sense, but `Dir('/opt')` above is not enough.

Xyene: The rules outlined in https://github.com/DMOJ/judge-server/issues/871 distinguish between these cases with `Self()` and `All()` atoms:

```python
    Dir("etc", (
        Self(),
        File("localtime"),
        File("timezone"),
    )),
    ...
    Dir("opt", (All(),))
```

dmoj/executors/filesystem_rules.py
Xyene: It never gave `/usr/`, t gave `/usr$` and `/usr/(?:home)`.

Xyene: Each rule has an implicit `^` at the start. The rule would have excluded home2, should such a user have existed. That's just an oversight.

dmoj/executors/filesystem_rules.py
Riolku: How exactly do we know `is_file`? Do I need to check for `O_PATH` or something? What about `stat` calls?

dmoj/cptbox/isolate.py
```diff
@@ -258,7 +263,7 @@ def _file_access_check(self, rel_file, debugger, is_open, flag_reg=1, dirfd=AT_F
 
         is_write = is_open and self.is_write_flags(getattr(debugger, 'uarg%d' % flag_reg))
         fs_jail = self.write_fs_jail if is_write else self.read_fs_jail
-        if fs_jail.match(file) is None:
+        if fs_jail.check(file) is not True:
```
kiritofeng, line 266:
```suggestion
        if not fs_jail.check(file):
```

dmoj/executors/SWIFT.py
```diff
@@ -3,9 +3,11 @@
 
 class Executor(CompiledExecutor):
     ext = 'swift'
+
```
kiritofeng, line 6: Do we actually want these newlines? (Or is this flake being silly)

kiritofeng (commented):
Will take another look tmr...

dmoj/executors/SWIFT.py
Riolku: Many of the newlines are me appeasing flake

dmoj/cptbox/isolate.py
Riolku: Lol, what am I doing

dmoj/executors/SWIFT.py
quantum5: Why would you need these lines? It makes no sense.

Riolku: Can we change something about a flake config option to remove the need for these lines?

quantum5: I am pretty sure it's not required except on your local machine, whose flake8 is somehow misconfigured.

Riolku: I installed flake8 with pip and ran `flake8` in the project root, I even tried maunally specifying `--config .flake8`... What would be wrong?

quantum5: Are you using the Python 2 version of flake8 or something? Because this should be impossible.

Riolku: No, I even tried re-installing with `pip3` and running with `python3 -m flake8`, but I'm still getting those errors. I could ignore them with `--ignore E302` but that won't solve the issue

kiritofeng: Removing the newlines and running `flake8` doesn't complain for me...

dmoj/executors/COFFEE.py
```diff
@@ -2,6 +2,8 @@
 
 from dmoj.executors.script_executor import ScriptExecutor
 
+from .filesystem_policies import FilesystemPolicy, ExactFile
```
Ninjaclasher, line 5: Do we want to use absolute imports?

Riolku: Unsure, defer to someone else

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def build(self):
+        self.root = Dir()
+
+        for rule in self.rules:
+            self.add_rule(rule)
+
+        self.built = True
+
+    def add_rule(self, rule : Rule):
```
Ninjaclasher, line 62: Will this method ever be called by the user outside of `build()`? If not, I think we usually prefix with `_` for clarity. If it will, perhaps we should add an assertion that `build()` was called so that you don't get a confusing `AttributeError` when `self.root` doesn't exist.

Riolku: I think in reality, the flow is "compose FilesystemPolicy objects into `fs` and then have the Tracer build it", in which case `build` could be prefixed with `_` as well.

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,109 @@
...
+        node = self.root
+
+        path_split = rule.path.split("/")
+
+        for index in range(len(path_split) - 1):
```
Ninjaclasher, line 67: We don't ever use `index` besides for indexing into `path_split`. Would it be more appropriate to iterate over the list directly? i.e. `for part in path_split[:-1]:`

(Similar comment for `check`).

dmoj/cptbox/isolate.py
```diff
@@ -185,11 +186,15 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
 
     def _compile_fs_jail(self, fs):
         if fs:
-            fs_re = '|'.join(fs)
+            fs.build()
+
+            return fs
         else:
-            fs_re = '(?!)'  # Disallow accessing everything by default.
+            dummy = FilesystemPolicy()
```
Ninjaclasher, line 193: Does this work? `FilesystemPolicy` is expecting an argument but we're not providing one here.

dmoj/executors/mono_executor.py
```diff
@@ -32,7 +34,10 @@ class MonoExecutor(CompiledExecutor):
     # get flagged as MLE.
     data_grace = 65536
     cptbox_popen_class = MonoTracedPopen
-    fs = ['/etc/mono/']
+    fs = FilesystemPolicy([
+        RecursiveDir('/etc/mono/')
```
Ninjaclasher, line 38: Does having a trailing slash here work? `add_rule` doesn't seem to strip it, so the last part of the path will be `''`. There are also some other fs directory rules that don't have a trailing slash and we should probably make them consistent.

dmoj/executors/filesystem_policies.py
Riolku: Maybe. This would require making a copy of the list for `add_rule`, which would impact speed. This is probably fine though. In `check`, there wouldn't be a performance hit, which is important.

dmoj/executors/mono_executor.py
Riolku: Trailing slashes need to be removed everywhere they appear, yeah.

dmoj/cptbox/isolate.py
Riolku: No, this will fail. It needs to be fixed.

Riolku: On a related note, is it preferable for FilesystemPolicy to use a variadic argument list or take a list?

dmoj/executors/COFFEE.py
Xyene: Yes, we do.

dmoj/cptbox/isolate.py
Xyene: Prefer named arguments over varargs for things that aren't typed by a human. Varargs mostly just make things more syntactically pleasing, at the cost of explicitness.

dmoj/executors/mono_executor.py
Xyene: We should assert that we have no empty components in the split path, to avoid reliance on human carefulness.

Riolku:
Java doesn't self-test properly, complaining about a Protection Fault on sys_fchdir, which is surprising, to say the least. The other tests pass though. I'm still unsure what to do about `is_file`, but otherwise I cleaned up the pending items.

Riolku:
Also, the flake8 issue is not resolved. Not sure what to do there either.

quantum5:
Temporary solution: ignore the local flake8 errors when you know they are wrong.

.freebsd.test.py
```diff
@@ -31,13 +33,16 @@
 
 def main():
     judgeenv.env['runtime'] = {}
-    judgeenv.env['extra_fs'] = {'PERL': ['/dev/dtrace/helper$'], 'RUBY2': ['/dev/dtrace/helper$']}
+    judgeenv.env['extra_fs'] = {
+        'PERL': FilesystemPolicy([ExactFile("/dev/dtrace/helper")]),
```
quantum5, line 37: This might be a problem: `extra_fs` is supposed to be passed from the yml, and will need to be of some format that YAML can represent.

dmoj/cptbox/isolate.py
```diff
@@ -258,9 +263,10 @@ def _file_access_check(self, rel_file, debugger, is_open, flag_reg=1, dirfd=AT_F
 
         is_write = is_open and self.is_write_flags(getattr(debugger, 'uarg%d' % flag_reg))
         fs_jail = self.write_fs_jail if is_write else self.read_fs_jail
-        if fs_jail.match(file) is None:
+        if fs_jail.check(file):
+            return file, True
+        else:
             return file, False
```
quantum5, line 269:
```suggestion
        return file, fs_jail.check(file)
```

dmoj/cptbox/isolate.py
```diff
@@ -185,11 +186,15 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
 
     def _compile_fs_jail(self, fs):
         if fs:
-            fs_re = '|'.join(fs)
+            fs.build()
+
```
quantum5, line 190: Excessive newlines, is this another flake8 madness? Starting to think it's something sketchy with line endings.

Riolku:
> Temporary solution: ignore the local flake8 errors when you know they are wrong.

I get E302 locally, but I see no ignore rule in `.flake8` for that...

quantum5:
> > Temporary solution: ignore the local flake8 errors when you know they are wrong.
> 
> I get E302 locally, but I see no ignore rule in `.flake8` for that...

E302 is for spacing between functions. Something must really be messed up. Do you have line-ending issues?

dmoj/cptbox/isolate.py
```diff
@@ -185,11 +186,15 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+            return fs
         else:
-            fs_re = '(?!)'  # Disallow accessing everything by default.
+            dummy = FilesystemPolicy([])
 
-        return re.compile(fs_re)
+            dummy.build()
+
```
quantum5, line 196: More excessive new lines.

dmoj/executors/COFFEE.py
```diff
@@ -2,6 +2,8 @@
 
 from dmoj.executors.script_executor import ScriptExecutor
```
quantum5, line 4: There shouldn't be any separation here. In general, no separation between DMOJ imports.

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,139 @@
...
+
+        self.map = {}
+
+
+class File:
+    pass
+
+
+class ExactFile:
+    def __init__(self, path : str):
```
quantum5, line 22: No space before `:`, in general.

Riolku:
> E302 is for spacing between functions. Something must really be messed up. Do you have line-ending issues?

```
riolku@midas:~/workspace/judge/judge$ xxd dmoj/executors/filesystem_policies.py 
00000000: 6672 6f6d 2074 7970 696e 6720 696d 706f  from typing impo
00000010: 7274 204c 6973 742c 2055 6e69 6f6e 0a0a  rt List, Union..
00000020: 0a63 6c61 7373 2041 4343 4553 535f 4d4f  .class ACCESS_MO
00000030: 4445 3a0a 2020 2020 4e4f 4e45 203d 2030  DE:.    NONE = 0
00000040: 0a20 2020 2045 5841 4354 203d 2031 0a20  .    EXACT = 1. 
00000050: 2020 2052 4543 5552 5349 5645 203d 2032     RECURSIVE = 2
00000060: 0a0a 0a63 6c61 7373 2044 6972 3a0a 2020  ...class Dir:.  
00000070: 2020 6465 6620 5f5f 696e 6974 5f5f 2873    def __init__(s
00000080: 656c 6629 3a0a 2020 2020 2020 2020 7365  elf):.        se
00000090: 6c66 2e61 6363 6573 735f 6d6f 6465 203d  lf.access_mode =
```

quantum5:
Then something is really messed up with your flake8 setup. What's the python version? Where are you running flake8?

The code as it currently is is not mergable. You have so many excessive new lines in your code that it hurts my eyes to read.

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,139 @@
...
+            self._add_rule(rule)
+
+        self.built = True
+
+    def _add_rule(self, rule : Rule):
+        if rule.path == "/":
+            return self._finalize_root_rule(rule)
+
+        node = self.root
+
```
quantum5, line 67: Get rid of all this double newline between code inside functions.

dmoj/executors/mixins.py
```diff
@@ -75,7 +115,11 @@ def get_security(self, launch_kwargs=None):
 
     def get_fs(self):
         name = self.get_executor_name()
-        fs = BASE_FILESYSTEM + self.fs + env.get('extra_fs', {}).get(name, []) + [re.escape(self._dir)]
+
```
quantum5, line 118: Destroy the empty line.

.freebsd.test.py
Riolku: What if we just use three keys, or change `extra_fs` to be a dictionary with three keys.

dmoj/cptbox/isolate.py
```diff
@@ -185,11 +184,9 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
 
     def _compile_fs_jail(self, fs):
         if fs:
```
quantum5, line 186: Could probably replace this entire function with `return FilesystemPolicy(fs or [])`.

.freebsd.test.py
```diff
@@ -31,13 +32,16 @@
 
 def main():
     judgeenv.env['runtime'] = {}
-    judgeenv.env['extra_fs'] = {'PERL': ['/dev/dtrace/helper$'], 'RUBY2': ['/dev/dtrace/helper$']}
+    judgeenv.env['extra_fs'] = {
+        'PERL': [ExactFile("/dev/dtrace/helper")],
+        'RUBY2': [ExactFile("/dev/dtrace/helper")]
```
quantum5, line 37: Maybe the YAML format could be something like:

```yaml
extra_fs:
  PERL:
  - file: /dev/dtrace/helper
  - dir: /some/exact/directory
  - recurse: /some/directory/and/all/children
```

quantum5 (commented):
Looks pretty good!

.freebsd.test.py
Riolku: sounds good to me, should I do that?

Xyene: Agree, though I'd prefer keeping the same names as the code (ExactDir, etc.)

Riolku: +1 on that. we could change the naming scheme to `exact_dir` and `recursive_dir`

quantum5: OK, use `exact_dir` and `recursive_dir` then.

.freebsd.test.py
```diff
@@ -31,13 +31,17 @@
 
 def main():
     judgeenv.env['runtime'] = {}
-    judgeenv.env['extra_fs'] = {'PERL': ['/dev/dtrace/helper$'], 'RUBY2': ['/dev/dtrace/helper$']}
+    judgeenv.env['extra_fs'] = {
+        'PERL': {"exact_file": ["/dev/dtrace/helper"]},
+        'RUBY2': {"exact_file": ["/dev/dtrace/helper"]}
```
quantum5, line 36: This is not what was agreed to earlier. The YAML we actually agreed to was:

```suggestion
        'PERL': [{"exact_file": "/dev/dtrace/helper"}],
        'RUBY2': [{"exact_file": "/dev/dtrace/helper"}],
```

Riolku: Why would it be a list of dictionaries

quantum5:
```yaml
extra_fs:
  PERL:
  - file: /dev/dtrace/helper
  - dir: /some/exact/directory
  - recurse: /some/directory/and/all/children
```

This is what we agreed to earlier. Clearly, `extra_fs.PERL` is a list of dictionaries, each with one key/value pair.

dmoj/executors/mixins.py
```diff
@@ -74,10 +114,27 @@ def get_security(self, launch_kwargs=None):
...
+        extra_fs = []
+        for rules in extra_fs_config:
+            for type, path in rules.iteritems():
+                if type == "exact_file":
+                    extra_fs.append(ExactFile(path))
+                elif type == "exact_dir":
+                    extra_fs.append(ExactDir(path))
+                else:
+                    assert type == "recursive_dir", f"Can't load rule for extra path with rule type {type}"
+                    extra_fs.append(RecursiveDir(path))
```
quantum5, line 134: Would probably be more pythonic to use dict-dispatch:

```python
rule_map = {'exact_file': ExactFile, ...}
...
constructor = rule_map[type]
assert constructor, f"Can't load rule for extra path with rule type {type}"
extra_fs.append(constructor(path))
```

quantum5 (commented):
Otherwise looks good.

quantum5 (approved):
Seems legit. Good job!

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,114 @@
+import os
+from typing import List, Union
+
+
+class ACCESS_MODE:
```
Xyene, line 5: Nit: should be `AccessMode` for consistency. Also, this should probably inherit from `Enum`, since it is one.

dmoj/executors/mixins.py
```diff
@@ -11,39 +11,79 @@
 from dmoj.judgeenv import env
 from dmoj.utils import setbufsize_path
 from dmoj.utils.unicode import utf8bytes
+from dmoj.executors.filesystem_policies import Rule, ExactDir, RecursiveDir, ExactFile
```
Xyene, line 14: Nit: dmoj imports should be alphabetically sorted. It is unfortunate flake is not enforcing this.

.freebsd.test.py
```diff
@@ -31,13 +31,18 @@
 
 def main():
     judgeenv.env['runtime'] = {}
-    judgeenv.env['extra_fs'] = {'PERL': ['/dev/dtrace/helper$'], 'RUBY2': ['/dev/dtrace/helper$']}
+    judgeenv.env['extra_fs'] = {
+        'PERL': [{"exact_file": "/dev/dtrace/helper"}],
```
Xyene, line 35: Nit: we consistently use `'` for strings except where we need to escape a `'`.

.freebsd.test.py
```diff
@@ -31,13 +31,18 @@
...
+        'RUBY2': [{"exact_file": "/dev/dtrace/helper"}]
```
Xyene, line 36: Nit: end this with a `,`, such that the next time a commit has to add a language here, only the new line will appear in the diff.

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,114 @@
...
+    NONE = 0
+    EXACT = 1
+    RECURSIVE = 2
+
+
+class Dir:
+    def __init__(self):
+        self.access_mode = ACCESS_MODE.NONE
+        self.map = {}
```
Xyene, line 14: Nit: what is this a map _of_? (It should be in the name.)

dmoj/executors/mixins.py
```diff
@@ -56,8 +96,8 @@ class PlatformExecutorMixin(metaclass=abc.ABCMeta):
     data_grace = 0
     fsize = 0
     personality = 0x0040000  # ADDR_NO_RANDOMIZE
-    fs: List[str] = []
-    write_fs: List[str] = []
+    fs: List[Rule] = []
```
Xyene, line 99: `Rule` is not descriptive here; it was fine inside `filesystem_policies.py`, but this is too far from that. Perhaps `Rule` should be renamed to something like `FilesystemAccessRule`.

dmoj/executors/mixins.py
```diff
@@ -74,10 +114,24 @@ def get_security(self, launch_kwargs=None):
         return self._add_syscalls(sec)
 
     def get_fs(self):
-        name = self.get_executor_name()
-        fs = BASE_FILESYSTEM + self.fs + env.get('extra_fs', {}).get(name, []) + [re.escape(self._dir)]
+        extra_fs = self._load_extra_fs()
+
```
Xyene, line 118: Nit: no extra newline necessary here. Honestly, the assignment to `fs` isn't, either.

dmoj/tests/test_filesystem_policy.py
```diff
@@ -0,0 +1,53 @@
+import unittest
```
Xyene, line 1: Some more tests:

- too many /s, e.g. `/etc/////passwd` and `////etc/passwd`
- empty string

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,114 @@
+import os
```
Xyene, line 1: This file should probably live inside `dmoj.cptbox` rather than `dmoj.executors`.

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,114 @@
...
+Rule = Union[ExactFile, ExactDir, RecursiveDir]
+
+
+class FilesystemPolicy:
+    def __init__(self, rules: List[Rule]):
+        self.root = Dir()
+        for rule in rules:
+            self._add_rule(rule)
+
+    def _add_rule(self, rule: Rule):
```
Xyene, line 49: Everything that has no return type should specify `-> None` as the type, otherwise mypy will usually infer the type as `Any`.

dmoj/executors/filesystem_policies.py
```diff
@@ -0,0 +1,114 @@
...
+            assert isinstance(new_node, File), "Can't add ExactFile: Dir rule exists"
+        else:
+            new_node = node.map.setdefault(final_component, Dir())
+            assert isinstance(new_node, Dir), "Can't add rule: File rule exists"
+            self._finalize_directory_rule(new_node, rule)
+
+    def _finalize_directory_rule(self, node: Dir, rule: Union[ExactDir, RecursiveDir]):
+        node.access_mode = max(node.access_mode, rule.access_mode)  # Allow the more permissive rule
+
+    # `path` should be a canonical path. The output of `realpath` is appropriate
```
Xyene, line 92: I think we ought to move the path normalization into `check`, since that avoids the class of bugs where we pass an un-normalized path to `check`. Until now it had to be elsewhere, since we were using regex directly and didn't have a wrapper class.

(This would be a good candidate for a unit test too.)

Xyene: Or maybe not, it's a bit annoying that we need the cwd for this.

Riolku: I'm not opposed to moving the logic in here, the CWD fetching is relatively straightforward, right? `check` could take `dir_fd` and `file_path`

dmoj/tests/test_filesystem_policy.py
Riolku: Should the rules themselves call `os.realpath` to avoid this madness? Or do we want this as a sanity check

dmoj/executors/filesystem_policies.py
Xyene: If `dir_fd` is `AT_FDCWD` you need to read the `cwd` of the calling process (https://github.com/DMOJ/judge-server/blob/e20c1d61b3a103fc80ddc5a12122b6b262a2c0a4/dmoj/cptbox/isolate.py#L268), so you need access to debugger instance, and have separate logic for Linux/FreeBSD to get the cwd.

Riolku: Yeah that's definitely annoying. We could pass the directory path and file path and have the filesystem policy handle joining

dmoj/executors/mixins.py
Riolku: https://github.com/tlocke/flake8-alphabetize can we make use of this?

dmoj/tests/test_filesystem_policy.py
Xyene: I think we could move the call to `os.path.normpath` from `isolate.py` into `check`. Then isolate just needs to worry about obtaining _some_ valid path, and we can normalize it here where it's unit-testable. @quantum5 wdyt?

dmoj/executors/mixins.py
Xyene: Probably https://github.com/PyCQA/flake8-import-order (by the simple metric of having more stars ⇒ more likely to stick around), but yeah. If you want to add this it'd be appreciated, but should be done in a separate PR.

dmoj/executors/filesystem_policies.py
Riolku: What functions/variables should have type annotations?

Riolku: And I guess, how much annotation is required/useful

Xyene: Everything that can be annotated should be annotated to the extent possible. https://github.com/DMOJ/judge-server/blob/master/dmoj/judge.py is a good example.

dmoj/tests/test_filesystem_policy.py
Riolku: Would we still need a check for an absolute path... probably, unless `check` takes two paths

quantum5: We could do `normpath`, yes, but what about conversion to absolute paths? That still needs to be handled by the caller. I guess we can have the caller do the join...

Also `os.path.realpath` resolves symlinks, and I am not sure if that's what we want. I think what we really want is to test both the original path and the path after resolving symlinks to avoid crazy attacks that way.

Xyene: Yes, let's just raise if a non-abspath is passed to `check`.

quantum5: On second thought, the only sane solution is to test both `normpath` and `realpath`. To deal with trickery, you can test to see if they are the same file with `samefile`.

dmoj/executors/filesystem_policies.py
Riolku: This will be resolved after #882 is merged.

Riolku:
Waiting on #872, and approval

Xyene:
Hmm, but #872 has already been merged...?

.freebsd.test.py
Xyene: Marked unresolved because this is the case for all other strings introduced in this PR. Again, I'm surprised flake8 didn't complain.

dmoj/cptbox/filesystem_policies.py
```diff
@@ -0,0 +1,113 @@
...
+class AccessMode(Enum):
+    NONE = 0
+    EXACT = 1
+    RECURSIVE = 2
+
+
+class Dir:
+    def __init__(self):
+        self.access_mode = AccessMode.NONE
+        self.subdirectory_map = {}
```
Xyene, line 15: Mmm, is this actually a map of subdirectories? My reading of the code suggests it can contain `File`s too.

dmoj/cptbox/filesystem_policies.py
```diff
@@ -0,0 +1,113 @@
...
+        if isinstance(rule, ExactFile):
+            new_node = node.subdirectory_map.setdefault(final_component, File())
+            assert isinstance(new_node, File), "Can't add ExactFile: Dir rule exists"
+        else:
+            new_node = node.subdirectory_map.setdefault(final_component, Dir())
+            assert isinstance(new_node, Dir), "Can't add rule: File rule exists"
+            self._finalize_directory_rule(new_node, rule)
+
+    def _finalize_directory_rule(self, node: Dir, rule: Union[ExactDir, RecursiveDir]) -> None:
+        node.access_mode = AccessMode(max(node.access_mode.value, rule.access_mode.value))  # Allow the more permissive rule
```
Xyene, line 90: We can move this into `AccessMode.most_permissive` to not leak the fact that these are ints here.

Xyene (commented):
Otherwise LGTM, thanks for working on this.

Riolku:
ran `black` and included changes you mentioned. I'll have to open a new PR with all the other formatting changes `black` wants to make.

Riolku:
Just realized I have a ton of quote issues, thanks to `flake8-quotes` I can actually see them now...

dmoj/cptbox/filesystem_policies.py
```diff
@@ -0,0 +1,117 @@
...
+
+class AccessMode(Enum):
+    NONE = 0
+    EXACT = 1
+    RECURSIVE = 2
+
+    # flake8 doesn't recognize `AccessMode` in type annotations here
+    @classmethod
+    def more_permissive(cls, mode1, mode2):
+        return AccessMode(max(mode1.value, mode2.value))
```
Ninjaclasher, line 14:
```suggestion
        return cls(max(mode1.value, mode2.value))
```
Is this more appropriate?

Riolku: I don't know. Maybe? Probably? I'll commit it anyway.

Riolku: Force pushed new commits that includes this change

.freebsd.test.py
Riolku: Fixed. #888 now lints this. Should be all clear.

Riolku:
Simply rebasing after #887

quantum5 (approved):

Riolku:
> This is the first step of migrating to Landlock support. A very simplistic implementation is given here.

Notably, this is no longer at all true. A trie implementation was taken instead.

PR #874, opened 2021-09-01
Detect WSL 1 and apply waitpid hack to it only

quantum5:
It appears that WSL 1 uses capitalized "Microsoft" while WSL2 doesn't,
and this is used to distinguish between the WSL versions.


Xyene (changes requested):
Please rename the define to be more explicit as well.

quantum5:
> Please rename the define to be more explicit as well.

Done.

Xyene (approved):
LGTM

hieplpvip:
Confirmed working on WSL2 Ubuntu 20.04

PR #875, opened 2021-09-04
testsuite: add test to make sure ENOENT is generated

quantum5:
`/dev/full` is is a device file that exists and can be opened by all users, but not actually useful and never actually used:

```console
$ ls -l /dev/full 
crw-rw-rw- 1 root root 1, 7 Aug 22 05:43 /dev/full
```


quantum5:
Removed hack for FreeBSD, test will pass after #877.

Riolku:
LGTM, FWIW

Xyene (approved):

PR #877, opened 2021-09-04
cptbox: use generic errno setter

quantum5:
This allows errno to be set correctly on FreeBSD.

### DIFF
--- a/dmoj/cptbox/ptdebug_freebsd_x64.cpp
+++ b/dmoj/cptbox/ptdebug_freebsd_x64.cpp
@@ -7,6 +7,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#define CARRY_FLAG 1
+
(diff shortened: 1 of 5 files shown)


dmoj/cptbox/ptdebug_freebsd_x64.cpp
```diff
@@ -7,6 +7,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#define CF 1
```
Xyene, line 10: Can we call this `CARRY_FLAG`?

Xyene (approved):
Otherwise LGTM.

dmoj/cptbox/ptdebug_freebsd_x64.cpp
quantum5: Done.

PR #878, opened 2021-09-04
ci: use python 3.9 for lint/mypy/sdist jobs

quantum5:
We no longer have Debian buster tethering us to the old Python 3.7, so let's upgrade.


kiritofeng (approved):
LGTM

PR #879, opened 2021-09-05
Return EPERM from file open functions

quantum5:
This is done for consistency with the planned landlock sandbox.


Xyene (approved):
LGTM, thanks

PR #880, opened 2021-09-05
testsuite: test overflow and invalid unicode for paths

quantum5:


Xyene (approved):
While we're at it, can we also add a test with a null path? We have a code path for that, but I don't _think_ we have a test for it?

Xyene (commented):
Actually, let's prefix all these tests with `sandbox_` so they sort together and we can instantly tell what tests we have/don't have in the future.

Xyene:
Am OK with leaving the rename to a separate PR, since we have other sandbox tests.

PR #882, opened 2021-09-05
cptbox: deal with symlink trickery

quantum5:
As discussed, checking both symlink itself and the symlink target, and also try to detect trickery by checking the inode.


dmoj/cptbox/isolate.py
```diff
@@ -247,26 +248,52 @@ def _file_access_check(self, rel_file, debugger, is_open, flag_reg=1, dirfd=AT_F
...
             log.exception('Unicode decoding error while opening relative to %d: %r', dirfd, rel_file)
-            return '(undecodable)', False
+            return '(undecodable)', ACCESS_EINVAL
 
-        is_write = is_open and self.is_write_flags(getattr(debugger, 'uarg%d' % flag_reg))
-        fs_jail = self.write_fs_jail if is_write else self.read_fs_jail
-        if fs_jail.match(file) is None:
-            return file, False
-        return file, True
+        actual = normalized = '/' + os.path.normpath(file).lstrip('/')
```
Xyene, line 262: Since this is a hack, let's at least leave a comment for why we're doing all this.

Xyene: Also, could we name the variables something a little more clear? We have `actual` `real` `normalized`, and the first two are synonyms of each other in English... I will confess to not being able to quickly grok this logic because of this.

quantum5: Fixed.

Xyene (approved):

PR #884, opened 2021-09-06
Make it an error to use bridged interactive graders on WSL 1

### LINKED ISSUE #863: Add warning if running bridged interactor on WSL
> WSL has broken `wait4` support, so you cannot debug two processes at one, and because the bridged interactor is also sandboxed, it won't work.

Thus we should raise an error if a judge running on WSL attempts to use a bridged interactor.

quantum5:
Fixes #863.


Xyene (approved):
LGTM

PR #886, opened 2021-09-06
cptbox: remove bit-rotten find_exe_in_path

quantum5:
There is no way `find_exe_in_path` actually works on Python 3 due to `str`/`bytes` confusion.


Xyene (approved):
LGTM, I don't think we ever actually used this anyway.

PR #887, opened 2021-09-06
apply `black` formatting to all of dmoj branch

Riolku:
This applies formatting changes by `black` to the entire branch. We should probably get formatting checks up with `flake8` for some of these things.


Xyene:
We use black 19.03 (see the workflow yml), for the avoidance of doubt do we still have diffs with that version?

Riolku:
Re-installed the exact version and re-ran black

quantum5 (approved):
Changes look reasonable.

Riolku:
https://pypi.org/project/flake8-black/ going to throw this in this PR later

quantum5:
ok to test

Riolku:
Added a flake8-black package to run checks. Also ran `black` against the testsuite, which definitely isn't necessary, and flake8 won't lint there, but the changes are probably still fine. Sidenote: why don't we lint the testsuite?

testsuite/bridged_checker/tests/py_ac/ac.py
```diff
@@ -1 +1 @@
-print -1, 1
+print - 1, 1
```
quantum5, line 1: Should we be running black on the testsuite? This is it attempting to apply Python 3 formatting on Python 2 code.

Riolku: Yikes... can we configure black somehow? Right now it takes a directory. We can apply to dmoj, but then there are some other files that deserve to be scanned... I'll do some research

Riolku: We have options to install `black[python2]`, but I think that's not the solution here.

quantum5: Definitely shouldn't be formatting the test suite. We could have deliberate syntax errors and other stuff. Should probably just exclude the entire directory from black and flake8.

Riolku: flake8 already excludes it. We can probably just run `black` with `--exclude testsuite`. I'll try that

Riolku: The command is `--exclude testsuite/`, since it seems to be a regex, it will ignore `testsuite.py`

Riolku: But happily, it will be caught by `flake8`.

quantum5 (commented):
Seems ok now. Maybe try putting an exclusion in `pyproject.toml` to avoid having to exclude the testsuite every time?

Riolku:
Should we be skipping string normalization with black? Seems we could benefit from it, if #888 is any indication

quantum5:
If it's fine post-#888 then we can remove the skip.

pyproject.toml
```diff
@@ -3,4 +3,5 @@ requires = ["setuptools", "wheel", "Cython"]
 
 [tool.black]
 line-length = 120
-skip-string-normalization = true
+skip-string-normalization = false
+exclude = '^/testsuite'
```
quantum5, line 7: Should probably be `extend-exclude` to avoid removing the default exclusions.

Riolku:
It is very not fine. It seems to prefer double quotes? What should we do? We can definitely let black run and do its thing, with the benefit of having it be able to auto-format for us, or we can stick to our current styling.

pyproject.toml
Riolku: That did not work at all for me locally. Not sure why. Is black outdated for us?

quantum5:
I'd rather have single quotes and enforce them with `flake8-quotes`.

pyproject.toml
Riolku: Works fine under 21.8b0 on Debian 11, but not 19 or whatever the version we have in the workflow is. What do we want to do?

quantum5: Yes. Try upgrading to `21.8b0` and see if it breaks anything. If not, let's just upgrade.

quantum5 (approved):

PR #888, opened 2021-09-06
add flake8-quotes and fix quoting issues

Riolku:
It seems `black` won't resolve these automatically. This also adds `flake8-quotes` to the workflow. https://pypi.org/project/flake8-quotes/


quantum5:
Where is the workflow change?

Riolku:
Fixed, thanks

quantum5 (approved):

quantum5:
ok to test

PR #893, opened 2021-09-07
cptbox: fix all compilation warnings

quantum5:
This PR is based on #892 and should be rebased after that one is merged.


Xyene (changes requested):
static inline is the preferred ordering, since GCC warns for inline static under some configurations.

Xyene (approved):
LGTM

PR #903, opened 2021-09-11
Initial implementation of landlock calls

Riolku:


kiritofeng:
ok to test

dmoj/cptbox/_cptbox.pyx
```diff
@@ -121,6 +125,13 @@ cdef extern from 'helper.h' nogil:
         bool use_seccomp
         int abi_for_seccomp
         int *seccomp_handlers
+        bool use_landlock
+        char** read_exact_files
```
Xyene, line 129: Nit: `char **read_exact_files`, here and elsewhere.

dmoj/cptbox/_cptbox.pyx
```diff
@@ -510,6 +521,16 @@ cdef class Process:
...
 
+            config.use_landlock = self._use_landlock()
+            if config.use_landlock:
+                config.read_exact_files = alloc_byte_array(self.read_exact_files)
+                config.read_exact_dirs = alloc_byte_array(self.read_exact_dirs)
+                config.read_recursive_dirs = alloc_byte_array(self.read_recursive_dirs)
+                config.write_exact_files = alloc_byte_array(self.write_exact_files)
+                config.write_exact_dirs = alloc_byte_array(self.write_exact_dirs)
+                config.write_recursive_dirs = alloc_byte_array(self.write_recursive_dirs)
+
```
Xyene, line 532: Nit: no double newline.

dmoj/cptbox/helper.cpp
```diff
@@ -117,6 +121,55 @@ int cptbox_child_run(const struct child_config *config) {
...
+    }
+
+#define READ_EXACT_FILE_RULE      LANDLOCK_ACCESS_FS_EXECUTE | LANDLOCK_ACCESS_FS_READ_FILE
+#define READ_EXACT_DIR_RULE       LANDLOCK_ACCESS_FS_READ_DIR
+#define READ_RECURSIVE_DIR_RULE   READ_EXACT_FILE_RULE | READ_EXACT_DIR_RULE
+#define WRITE_EXACT_FILE_RULE     LANDLOCK_ACCESS_FS_WRITE_FILE
+#define WRITE_EXACT_DIR_RULE      LANDLOCK_ACCESS_FS_READ_DIR
+#define WRITE_RECURSIVE_DIR_RULE  WRITE_EXACT_FILE_RULE | WRITE_EXACT_DIR_RULE
+
+    if(
```
Xyene, line 156: Nit: `if (`

dmoj/cptbox/landlock_helper.c
```diff
@@ -0,0 +1,58 @@
+#include "landlock_helpers.h"
+
+#if PTBOX_LANDLOCK
+    #include <linux/landlock.h>
+
+    int add_rules(const int ruleset_fd, const char* const* const paths, __u64 access_rule) {
```
Xyene, line 6:
```suggestion
    int landlock_add_rules(const int ruleset_fd, const char* const* const paths, __u64 access_rule) {
```

dmoj/cptbox/landlock_helper.c
```diff
@@ -0,0 +1,58 @@
...
+            if(path_beneath.parent_fd < 0) {
+                perror("Failed to open path for rule");
+                return -1;
+            }
+            if(landlock_add_rule(ruleset_fd, LANDLOCK_RULE_PATH_BENEATH, &path_beneath, 0)) {
+                perror("Failed to add rule to ruleset");
+                return -1;
+            }
+            if(close(path_beneath.parent_fd)) {
+                // Not Fatal: we have a CLOEXEC flag
```
Xyene, line 23: Don't worry about logging this.

dmoj/cptbox/landlock_helper.c
```diff
@@ -0,0 +1,58 @@
...
+#if PTBOX_LANDLOCK
+    #include <linux/landlock.h>
+
+    int add_rules(const int ruleset_fd, const char* const* const paths, __u64 access_rule) {
+        struct landlock_path_beneath_attr path_beneath = {
+            .parent_fd = -1,
+            .allowed_access = access_rule,
+        };
+
+        for(const char* const* pathptr = paths; *pathptr; pathptr++) {
```
Xyene, line 12: Where is this array being null-terminated? Is this relying on `alloc_byte_array` ending in a null?

dmoj/cptbox/landlock_helper.c
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        return 0;
+    }
+
+    #ifndef landlock_create_ruleset
+    int landlock_create_ruleset(
+    		const struct landlock_ruleset_attr *const attr,
+    		const size_t size, const __u32 flags)
+    {
```
Xyene, line 35: Nit: `{` on previous line, here and elsewhere.

dmoj/cptbox/landlock_helper.c
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+
+        return 0;
+    }
+
+    #ifndef landlock_create_ruleset
+    int landlock_create_ruleset(
+    		const struct landlock_ruleset_attr *const attr,
+    		const size_t size, const __u32 flags)
```
Xyene, line 34: Nit: attempt fitting as many params on one 80-col line before wrapping.

dmoj/cptbox/helper.cpp
```diff
@@ -117,6 +121,55 @@ int cptbox_child_run(const struct child_config *config) {
...
+        add_rules(ruleset_fd, config->read_recursive_dirs,  READ_RECURSIVE_DIR_RULE) ||
+        add_rules(ruleset_fd, config->write_exact_files,    WRITE_EXACT_FILE_RULE) ||
+        add_rules(ruleset_fd, config->write_exact_dirs,     WRITE_EXACT_DIR_RULE) ||
+        add_rules(ruleset_fd, config->write_recursive_dirs, WRITE_RECURSIVE_DIR_RULE)
+    ) {
+        return PTBOX_SPAWN_FAIL_LANDLOCK;
+    }
+
+    if(landlock_restrict_self(ruleset_fd, 0)) {
+    		perror("Failed to enforce ruleset");
```
Xyene, line 168: Nit: inconsistent tabs and spaces, here and in a lot of other places. We use 4-space tabs.

dmoj/cptbox/ptproc.cpp
```diff
@@ -105,6 +105,15 @@ bool pt_process::use_seccomp(bool enabled) {
     return true;
 }
 
+bool pt_process::use_landlock(bool enabled) {
```
Xyene, line 108: Is the return value of this function supposed to be if turning landlock on was successful? If so, it should return `false` if we couldn't turn it on because we don't have landlock support.

dmoj/cptbox/helper.h
```diff
@@ -6,6 +6,7 @@
 #define PTBOX_SPAWN_FAIL_SECCOMP        203
 #define PTBOX_SPAWN_FAIL_TRACEME        204
 #define PTBOX_SPAWN_FAIL_EXECVE         205
+#define PTBOX_SPAWN_FAIL_LANDLOCK       206
```
Xyene, line 9: Needs some new handling code in https://github.com/DMOJ/judge-server/blob/eefbcbba4194fea472ef51287396b38682af24cc/dmoj/cptbox/tracer.py#L207-L221 too

dmoj/cptbox/ptbox.h
```diff
@@ -29,6 +29,12 @@
 #include <seccomp.h>
 #endif
 
+#ifdef PTBOX_USE_LANDLOCK
```
Xyene, line 32: Thinking more about it, we likely do need runtime detection of Landlock capabilities.

With Landlock, we need to pass in `ALLOW` to seccomp for all syscalls that are handled by Landlock, so we don't go through both a Landlock check _and_ a ptrace check.

We can detect the presence of Landlock at runtime by issuing `sys_landlock_create_ruleset` and checking that the errno is not `ENOSYS`.

Riolku: Alright, sounds good

quantum5: I was thinking of having a list of system call numbers prepared for landlock, and allowing those if landlock creation succeeds. This would likely result in additional plumbing, but has the advantage of doing detection at the exact spot where it's used.

Xyene: The downside of that is it's a little surprising to have handlers listed in `isolate.py` that never end up getting called. I don't think specifying a handler that's also handled by landlock should be disallowed (e.g., might want to do so just to log all filesystem accesses), it's just not particularly useful.

dmoj/cptbox/landlock_helper.c
Riolku: Yes

dmoj/cptbox/ptbox.h
Riolku: We'd want to mark handlers as conditional on landlock or something then, and export that down to the landlock setup code

Riolku:
Landlock applies `LANDLOCK_FS_READ_DIR` recursively. This means that since we allow `READ_DIR /`, a process can search the whole filesystem tree. We can limit `getdents`, but a process can still arbitrarily determine the existence of a file/directory. This might be acceptable.

Riolku:
Given that stuff is now functional, we can begin to review this.

dmoj/cptbox/_cptbox.pyx
```diff
@@ -13,7 +13,7 @@ __all__ = ['Process', 'Debugger', 'bsd_get_proc_cwd', 'bsd_get_proc_fdno', 'MAX_
            'PTBOX_ABI_X86', 'PTBOX_ABI_X64', 'PTBOX_ABI_X32', 'PTBOX_ABI_ARM', 'PTBOX_ABI_ARM64',
            'PTBOX_ABI_FREEBSD_X64', 'PTBOX_ABI_INVALID', 'PTBOX_ABI_COUNT',
            'PTBOX_SPAWN_FAIL_NO_NEW_PRIVS', 'PTBOX_SPAWN_FAIL_SECCOMP', 'PTBOX_SPAWN_FAIL_TRACEME',
-           'PTBOX_SPAWN_FAIL_EXECVE']
+           'PTBOX_SPAWN_FAIL_EXECVE', 'PTBOX_SPAWN_FAIL_LANDLOCK']
```
quantum5, line 16: What about `has_landlock`?

dmoj/cptbox/_cptbox.pyx
```diff
@@ -218,6 +226,12 @@ def memory_fd_seal(int fd):
     if result == -1:
         PyErr_SetFromErrno(OSError)
 
+def has_landlock():
+    cdef int code = has_landlock_check();
```
quantum5, line 230:
```suggestion
    cdef int code = has_landlock_check()
```

dmoj/cptbox/helper.cpp
```diff
@@ -31,6 +32,7 @@
 #   define FD_DIR "/proc/self/fd"
 #endif
 
+
```
quantum5, line 35:
```suggestion

```

dmoj/cptbox/helper.cpp
```diff
@@ -73,13 +75,69 @@ int cptbox_child_run(const struct child_config *config) {
...
+    };
+
+    ruleset_fd =
+        landlock_create_ruleset(&ruleset_attr, sizeof(ruleset_attr), 0);
+    if (ruleset_fd < 0) {
+        if (errno != ENOSYS && errno != EOPNOTSUPP) {
+            perror("Failed to create a ruleset");
+            return PTBOX_SPAWN_FAIL_LANDLOCK;
+        } else
+            goto seccomp_setup;
```
quantum5, line 99:
```suggestion
        }
        goto seccomp_setup;
```

dmoj/cptbox/helper.cpp
```diff
@@ -73,13 +75,69 @@ int cptbox_child_run(const struct child_config *config) {
...
+    }
+
+#define READ_EXACT_FILE_RULE                                                   \
+    LANDLOCK_ACCESS_FS_EXECUTE | LANDLOCK_ACCESS_FS_READ_FILE
+#define READ_EXACT_DIR_RULE LANDLOCK_ACCESS_FS_READ_DIR
+#define READ_RECURSIVE_DIR_RULE READ_EXACT_FILE_RULE | READ_EXACT_DIR_RULE
+#define WRITE_EXACT_FILE_RULE LANDLOCK_ACCESS_FS_WRITE_FILE
+#define WRITE_EXACT_DIR_RULE LANDLOCK_ACCESS_FS_READ_DIR
+#define WRITE_RECURSIVE_DIR_RULE WRITE_EXACT_FILE_RULE | WRITE_EXACT_DIR_RULE
```
quantum5, line 108: Do we need these macros? They are used exactly once.

dmoj/cptbox/helper.h
```diff
@@ -22,8 +23,16 @@ struct child_config {
     int stdout_;
     int stderr_;
     int *seccomp_handlers;
+    char** read_exact_files;
+    char** read_exact_dirs;
+    char** read_recursive_dirs;
+    char** write_exact_files;
+    char** write_exact_dirs;
+    char** write_recursive_dirs;
```
quantum5, line 31: Pointer star on the wrong side.

dmoj/cptbox/helper.cpp
```diff
@@ -141,9 +199,25 @@ int cptbox_child_run(const struct child_config *config) {
...
+        if(errno == ENOSYS || errno == EOPNOTSUPP) {
+            return 0;
+        }
+        else {
+            return -1;
+        }
+    }
+    else {
+        return 1;
+    }
```
quantum5, line 217: I hate this control flow.

```suggestion
    int landlock_version = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
    if (landlock_version >= 0)
        return 1;
    if (errno == ENOSYS || errno == EOPNOTSUPP)
        return 0;
    return -1;
```

dmoj/cptbox/landlock_header.h
```diff
@@ -0,0 +1,76 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+/*
+ * Landlock - User space API
+ *
+ * Copyright © 2017-2020 Mickaël Salaün <mic@digikod.net>
+ * Copyright © 2018-2020 ANSSI
+ */
```
quantum5, line 7: No include guard around this header.

dmoj/cptbox/landlock_helpers.cpp
```diff
@@ -0,0 +1,37 @@
+#ifndef __FreeBSD__
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+
+#include "landlock_helpers.h"
+
+int landlock_add_rules(const int ruleset_fd, const char *const *const paths,
```
quantum5, line 10: `const char **paths` would have been fine. The amount of `const`s you are adding is excessive.

dmoj/cptbox/landlock_helpers.cpp
```diff
@@ -0,0 +1,37 @@
...
+                       __u64 access_rule) {
+    struct landlock_path_beneath_attr path_beneath = {
+        .allowed_access = access_rule,
+        .parent_fd = -1,
+    };
+    for (const char *const *pathptr = paths; *pathptr; pathptr++) {
```
quantum5, line 16: No need for the second `const` after previous change.

dmoj/cptbox/tracer.py
```diff
@@ -198,6 +202,17 @@ def _get_seccomp_handlers(self) -> List[int]:
                     handlers[call] = handler.errno
         return handlers
 
+    def load_files(self, read_fs, write_fs) -> None:
+        def filter(source: List[FilesystemAccessRule], type) -> List[bytes]:
```
quantum5, line 206: Don't trick other people by reusing builtin function names.

dmoj/cptbox/isolate.py
```diff
@@ -44,11 +44,19 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
             self._getcwd_pid = lambda pid: os.readlink('/proc/%d/cwd' % pid)
             self._getfd_pid = lambda pid, fd: os.readlink('/proc/%d/fd/%d' % (pid, fd))
 
+        if has_landlock():
+            self.update({sys_openat: ALLOW, sys_open: ALLOW})
```
quantum5, line 48: Probably best to implement a debug mode that uses a callback to log all the `EACCES` generated by landlock. This will be highly useful for debugging executor failures.

Riolku: How would debug mode be set?

quantum5: Constructor argument, set to true during self-test.

dmoj/cptbox/helper.cpp
Riolku: If anything they serve clarity, and you can see that `READ_RECURSIVE_DIR_RULE` is composed of two smaller macros, so there could be some value.

Riolku: I don't mind either way if they are removed or not, up to you.

dmoj/cptbox/helper.cpp
```diff
@@ -73,13 +74,61 @@ int cptbox_child_run(const struct child_config *config) {
...
+            LANDLOCK_ACCESS_FS_READ_FILE | LANDLOCK_ACCESS_FS_READ_DIR |
+            LANDLOCK_ACCESS_FS_REMOVE_DIR | LANDLOCK_ACCESS_FS_REMOVE_FILE |
+            LANDLOCK_ACCESS_FS_MAKE_CHAR | LANDLOCK_ACCESS_FS_MAKE_DIR |
+            LANDLOCK_ACCESS_FS_MAKE_REG | LANDLOCK_ACCESS_FS_MAKE_SOCK |
+            LANDLOCK_ACCESS_FS_MAKE_FIFO | LANDLOCK_ACCESS_FS_MAKE_BLOCK |
+            LANDLOCK_ACCESS_FS_MAKE_SYM,
+    };
+
+    ruleset_fd =
+        landlock_create_ruleset(&ruleset_attr, sizeof(ruleset_attr), 0);
```
quantum5, line 92:
```suggestion
    ruleset_fd = landlock_create_ruleset(&ruleset_attr, sizeof(ruleset_attr), 0);
```

dmoj/cptbox/tracer.py
```diff
@@ -198,6 +202,17 @@ def _get_seccomp_handlers(self) -> List[int]:
                     handlers[call] = handler.errno
         return handlers
 
+    def load_files(self, read_fs, write_fs) -> None:
```
quantum5, line 205: Missing annotations on `read_fs` and `write_fs`.

dmoj/cptbox/tracer.py
```diff
@@ -198,6 +202,17 @@ def _get_seccomp_handlers(self) -> List[int]:
...
+        def _get_rule_paths(source: List[FilesystemAccessRule], type) -> List[bytes]:
```
quantum5, line 206: Missing annotations on `type`.

dmoj/cptbox/isolate.py
```diff
@@ -44,11 +45,24 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
             self._getcwd_pid = lambda pid: os.readlink('/proc/%d/cwd' % pid)
             self._getfd_pid = lambda pid, fd: os.readlink('/proc/%d/fd/%d' % (pid, fd))
 
+        if has_landlock():
+            if debug_landlock:
+                self.update(
+                    {sys_openat: self.debug_file_access_at('openat'), sys_open: self.debug_file_access('open', 0)}
```
quantum5, line 51: Are these the only calls landlock restricts? You might want to add more...

dmoj/cptbox/isolate.py
```diff
@@ -217,6 +231,41 @@ def read_path(syscall: str, debugger: Debugger, ptr: int):
...
+
+    def _debug_file_access(self, rel_file, debugger, dirfd=AT_FDCWD) -> None:
+        try:
+            file = self.get_full_path(debugger, rel_file, dirfd)
+        except UnicodeDecodeError:
+            log.debug('Unicode decoding error when debugging file access relative to %d: %r', dirfd, rel_file)
+            return
+
+        def on_return() -> None:
+            if debugger.result == -errno.EACCES:
```
quantum5, line 264:
```suggestion
            if debugger.errno == errno.EACCES:
```

dmoj/cptbox/isolate.py
Riolku: Landlock doesn't restrict a lot of calls right now. From the [docs](https://www.kernel.org/doc/html/latest/userspace-api/landlock.html#landlock-rules):

> It is currently not possible to restrict some file-related actions accessible through these syscall families: chdir(2), truncate(2), stat(2), flock(2), chmod(2), chown(2), setxattr(2), utime(2), ioctl(2), fcntl(2), access(2). Future Landlock evolutions will enable to restrict them.

This leaves us with... basically nothing. Most of our file access calls are variants on `stat`, `access`, and `readlink`, none of which can be restricted. Notably, `readlink` is not considered a "file-related" action, as far as I understand: landlock hooks in _after_ the resolution is complete, including symlinks.

I wish it were more.

Riolku: I guess, if nothing else; `open` is probably the most important syscall set? I guess I don't know what the syscall counts look like.

quantum5 (approved):
Don't see anything particularly wrong. @Xyene should probably also take a look.

dmoj/cptbox/compiler_isolate.py
```diff
@@ -103,6 +148,47 @@ def __init__(self, *, tmpdir, read_fs, write_fs):
...
+
+    @syscall_simulator
+    def simulate_link(self, debugger: Debugger) -> None:
+        dst = self.fullpath_from_reg_and_dirfd(debugger, file_reg=1, dirfd=AT_FDCWD)
+        self._access_check(debugger, dst, self.write_fs_jail)
+        src = self.fullpath_from_reg_and_dirfd(debugger, file_reg=0, dirfd=AT_FDCWD)
+        os.link(src, dst)
+
+    @syscall_simulator
+    def simulate_linkat(self, debugger: Debugger) -> None:
```
Xyene, line 163: I'd feel a lot more confident about these *at calls if we validated that the `dirfd` is something we know about, and disallowed otherwise. It's easy to imagine allowing a call we shouldn't have because we got the `at` part wrong. This was less scary when the most that could happen was reads.

Xyene: Sorry, I mean we should probably be validating the flags register.

Riolku: So, throw the dirfd through the jail as well?

Riolku: Like, read the path given by `dirfd`, and validate that they have at least read perms on it, and then check that the full path has write perms?

Xyene: No, I mean that for instance `linkat` has a flag register that can specify `AT_SYMLINK_FOLLOW`, which changes how the path should be interpreted. We don't implement support for this flag. As long as that is true, we shouldn't allow that flag. Failing with `EINVAL` if flags is nonzero is safe.

Riolku: Do we need to do this for generic checking too (like in handle_renameat), and if so, for which calls?

Riolku: Added flag checking for `linkat`, since that is the only call with flags that we simulate. `renameat` does not have flags; `renameat2` does.

Riolku:
It seems that my changes to the handling of flags were a figment of my imagination. They should be present as needed now.

Xyene (commented):
I think we should log at startup whether we're using:

* `ptrace`
* `seccomp + ptrace`
* `landlock + seccomp + ptrace`

sandboxing modes. Otherwise it seems a little difficult to verify we've actually deployed Landlock correctly.

dmoj/cptbox/compiler_isolate.py
```diff
@@ -13,6 +16,25 @@
 UTIME_OMIT = (1 << 30) - 2
 
 
+def syscall_simulator(
+    simulator: Callable[['CompilerIsolateTracer', Debugger], None]
+) -> Callable[['CompilerIsolateTracer', Debugger], None]:
+    def inner(tracer: 'CompilerIsolateTracer', debugger: Debugger) -> None:
```
Ninjaclasher, line 22: Should we `@functools.wraps(simulator)`?

Riolku: Unfamiliar with that function. I can look into it (or blindly follow your suggestion).

dmoj/cptbox/compiler_isolate.py
```diff
@@ -103,6 +149,51 @@ def __init__(self, *, tmpdir, read_fs, write_fs):
...
+    def fullpath_from_reg_and_dirfd(self, debugger: Debugger, *, file_reg, dirfd):
+        rel_file = self.get_rel_file(debugger, reg=file_reg)
+        return self.get_full_path_unnormalized(debugger, rel_file, dirfd=dirfd)
+
+    @syscall_simulator
+    def simulate_link(self, debugger: Debugger) -> None:
+        dst = self.fullpath_from_reg_and_dirfd(debugger, file_reg=1, dirfd=AT_FDCWD)
+        self._access_check(debugger, dst, self.write_fs_jail)
+        src = self.fullpath_from_reg_and_dirfd(debugger, file_reg=0, dirfd=AT_FDCWD)
+        os.link(src, dst)
```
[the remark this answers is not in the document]
Riolku: `link` source isn't checked under seccomp right now. I feel like either way we go, we should have consistency here. What do you think?

dmoj/cptbox/compiler_isolate.py
```diff
@@ -13,6 +16,25 @@
...
 
 
+def syscall_simulator(
+    simulator: Callable[['CompilerIsolateTracer', Debugger], None]
+) -> Callable[['CompilerIsolateTracer', Debugger], None]:
+    def inner(tracer: 'CompilerIsolateTracer', debugger: Debugger) -> None:
+        try:
+            simulator(tracer, debugger)
+            code = 0
+        except OSError as e:
```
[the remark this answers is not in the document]
Riolku: I'd expect to get a protection fault, and I think that'd be ideal. If something other than `OSError` is thrown, what would we do?

Riolku:
> I think we should log at startup whether we're using:
> 
>     * `ptrace`
> 
>     * `seccomp + ptrace`
> 
>     * `landlock + seccomp + ptrace`
> 
> 
> sandboxing modes. Otherwise it seems a little difficult to verify we've actually deployed Landlock correctly.

Where should this be done? In `dmoj/judge.py`?

dmoj/cptbox/compiler_isolate.py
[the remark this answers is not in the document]
Riolku: Let's check this here, and make a separate PR to add this to seccomp at a later time.

dmoj/cptbox/compiler_isolate.py
Riolku: Added.

dmoj/cptbox/compiler_isolate.py
```diff
@@ -13,6 +16,25 @@
 UTIME_OMIT = (1 << 30) - 2
 
 
+def syscall_simulator(
+    simulator: Callable[['CompilerIsolateTracer', Debugger], None]
+) -> Callable[['CompilerIsolateTracer', Debugger], None]:
+    def inner(tracer: 'CompilerIsolateTracer', debugger: Debugger) -> None:
+        try:
+            simulator(tracer, debugger)
+            code = 0
```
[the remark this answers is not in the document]
Riolku: Done.

dmoj/cptbox/compiler_isolate.py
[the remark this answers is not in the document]
Riolku: Resolving.

Riolku:
Startup message added.

Xyene:
Landlock ABI 3 shipped in 5.19 which I think is able to supplant our syscall simulation here.

We should hard-require it; @quantum5 is working on making a 6.x kernel available in the ARM64 CI runner.

TODO:
- [ ] Read up on `LANDLOCK_ACCESS_FS_REFER`. [Slides.](https://static.sched.com/hosted_files/lssna2022/b7/2022-06-24%20LSS%20-%20Update%20on%20Landlock.pdf) It seems like we could refer to `/` and things might "Just Work" as we've set them up?
- [ ] Address the truncation comments in https://docs.kernel.org/next/userspace-api/landlock.html#truncating-files.

Riolku:
> Landlock ABI 3 shipped in 5.19 which I think is able to supplant our syscall simulation here.
> 
> We should hard-require it; @quantum5 is working on making a 6.x kernel available in the ARM64 CI runner.

Would this also mean we can do away with some of the header files that we copied? `landlock_header.h` specifically.

Xyene:
No, we need them to be able to do the initial Landlock calls to figure out if Landlock is supported at all.

dmoj/cptbox/helper.cpp
```diff
@@ -90,13 +91,62 @@ int cptbox_child_run(const struct child_config *config) {
...
+    struct landlock_ruleset_attr ruleset_attr = {
+        .handled_access_fs = LANDLOCK_ACCESS_FS_EXECUTE | LANDLOCK_ACCESS_FS_WRITE_FILE | LANDLOCK_ACCESS_FS_READ_FILE |
+                             LANDLOCK_ACCESS_FS_READ_DIR | LANDLOCK_ACCESS_FS_REMOVE_DIR |
+                             LANDLOCK_ACCESS_FS_REMOVE_FILE | LANDLOCK_ACCESS_FS_MAKE_REG |
+                             LANDLOCK_ACCESS_FS_MAKE_SYM | LANDLOCK_ACCESS_FS_MAKE_DIR | LANDLOCK_ACCESS_FS_REFER,
+    };
+
+    int landlock_version = get_landlock_version();
+    if (landlock_version < 2) {
+        // ABI not old enough. Skip to seccomp
```
Xyene, line 105:
```suggestion
        // ABI not new enough. Skip to seccomp
```

dmoj/cptbox/helper.cpp
```diff
@@ -90,13 +91,62 @@ int cptbox_child_run(const struct child_config *config) {
...
+        landlock_add_rules(ruleset_fd, config->landlock_write_exact_files,
+                           LANDLOCK_ACCESS_FS_READ_FILE | LANDLOCK_ACCESS_FS_EXECUTE | LANDLOCK_ACCESS_FS_WRITE_FILE) ||
+        landlock_add_rules(ruleset_fd, config->landlock_write_exact_dirs, LANDLOCK_ACCESS_FS_READ_DIR) ||
+        landlock_add_rules(ruleset_fd, config->landlock_write_recursive_dirs,
+                           LANDLOCK_ACCESS_FS_READ_FILE | LANDLOCK_ACCESS_FS_EXECUTE | LANDLOCK_ACCESS_FS_READ_DIR |
+                               LANDLOCK_ACCESS_FS_WRITE_FILE | LANDLOCK_ACCESS_FS_REMOVE_DIR |
+                               LANDLOCK_ACCESS_FS_REMOVE_FILE | LANDLOCK_ACCESS_FS_READ_DIR |
+                               LANDLOCK_ACCESS_FS_MAKE_REG | LANDLOCK_ACCESS_FS_MAKE_SYM | LANDLOCK_ACCESS_FS_MAKE_DIR |
+                               LANDLOCK_ACCESS_FS_REFER)) {
+        // landlock_add_rules logs errors
```
Xyene, line 131: Could this comment have more details? I don't understand what it's referring to, is this comment unrelated to the following `close`?

dmoj/cptbox/helper.cpp
```diff
@@ -175,6 +225,26 @@ int cptbox_child_run(const struct child_config *config) {
...
+
+    int landlock_version = landlock_create_ruleset(NULL, 0, LANDLOCK_CREATE_RULESET_VERSION);
+    if (landlock_version >= 0) {
+        return landlock_version;
+    }
+    if (errno == ENOSYS || errno == EOPNOTSUPP)
+        return 0;
+    return -1;
+#else
+    return 0;  // FreeBSD does not have landlock
```
Xyene, line 244: We should `abort()` here.

dmoj/cptbox/landlock_header.h
```diff
@@ -0,0 +1,76 @@
...
+#define LANDLOCK_ACCESS_FS_MAKE_CHAR    (1ULL << 6)
+#define LANDLOCK_ACCESS_FS_MAKE_DIR     (1ULL << 7)
+#define LANDLOCK_ACCESS_FS_MAKE_REG     (1ULL << 8)
+#define LANDLOCK_ACCESS_FS_MAKE_SOCK    (1ULL << 9)
+#define LANDLOCK_ACCESS_FS_MAKE_FIFO    (1ULL << 10)
+#define LANDLOCK_ACCESS_FS_MAKE_BLOCK   (1ULL << 11)
+#define LANDLOCK_ACCESS_FS_MAKE_SYM     (1ULL << 12)
+#endif /* _LINUX_LANDLOCK_H */
+
+// Not always defined, depends on ABI version.
```
Xyene, line 44: `ifndef`?

dmoj/cptbox/landlock_helpers.h
```diff
@@ -0,0 +1,3 @@
+#include "landlock_header.h"
+
+int landlock_add_rules(const int ruleset_fd, const char **paths, __u64 access_rule);
```
Xyene, line 3: Could we name this `cptbox_landlock_add_rules`? Makes it easier to keep track of what's provided by the kernel versus by us.

dmoj/cptbox/landlock_helpers.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <string>
+
+#include "landlock_helpers.h"
+
+int landlock_add_path(struct landlock_path_beneath_attr &rule, const int ruleset_fd, const char *path) {
+    rule.parent_fd = open(path, O_PATH | O_CLOEXEC);
+
+    if (rule.parent_fd < 0) {
+        if (errno == ENOENT)
+            goto close_fd;  // missing files are ignored
```
Xyene, line 16: Nit: parens even around single-statement branches / loops / etc. I know not all cptbox code is written like this, but all new code should be.

dmoj/cptbox/tracer.py
```diff
@@ -149,6 +150,10 @@ def __init__(
         self.protection_fault = None
 
         self._security = security
+        if security is not None:
+            self.configure_files(security.read_fs, security.write_fs)
+        else:
+            self.configure_files([], [])
```
Xyene, line 156: Is this branch necessary at all? Seems a little weird.

dmoj/cptbox/tracer.py
```diff
@@ -223,6 +244,8 @@ def wait(self) -> int:
                 raise RuntimeError('failed to spawn child')
             elif self.returncode == PTBOX_SPAWN_FAIL_SETAFFINITY:
                 raise RuntimeError('failed to set child affinity')
+            elif self.returncode == PTBOX_SPAWN_FAIL_LANDLOCK:
+                raise RuntimeError('landlock configuration failed')
```
Xyene, line 248:
```suggestion
                raise RuntimeError('Landlock configuration failed')
```

Nit, here and in other user-facing strings as well as comments. The kernel docs refer to it as "Landlock" (despite "seccomp" being lowercased).

dmoj/executors/SCALA.py
```diff
@@ -20,6 +20,7 @@ class Executor(JavaExecutor):
         ExactFile('/bin/bash'),
         RecursiveDir('/etc/alternatives'),
     ]
+    compiler_syscalls = ['mincore']
```
Xyene, line 23: Why did this change?

dmoj/cptbox/helper.cpp
```diff
@@ -175,6 +225,26 @@ int cptbox_child_run(const struct child_config *config) {
 #endif
 }
 
+int get_landlock_version() {
+#if !PTBOX_FREEBSD
+    char *sandbox_mode = getenv("DMOJ_SANDBOX_MODE");
+    if (sandbox_mode != nullptr && strcmp(sandbox_mode, "ptrace+seccomp") == 0) {
```
Xyene, line 231: This is a surprising place for this check to live, and I think the check should be stricter than this.

We should validate that the string is one of:

* `ptrace+seccomp`
* `ptrace+seccomp+landlock`
* `auto`

and otherwise bail out. Otherwise we can't confidently make changes to the interpretation of this environment variable, as we'd have been too lax in its inputs in earlier versions.

dmoj/cptbox/helper.h
```diff
@@ -27,8 +28,16 @@ struct child_config {
     int *seccomp_handlers;
     // 64 cores ought to be enough for anyone.
     unsigned long cpu_affinity_mask;
+    const char **landlock_read_exact_files;
```
Xyene, line 31: This may end up looking a little cleaner if we have a

```c
struct {
    const char **read_exact_files;
    ...
} landlock;
```

or even

```c
const struct {
    char **read_exact_files;
    ...
} landlock;
```

dmoj/executors/RUST.py
```diff
@@ -86,6 +87,12 @@ def get_shared_target(self):
                 # We intentionally don't clean this directory up at any point, since we can re-use it.
                 return self.shared_target
 
+    def get_fs(self) -> List[FilesystemAccessRule]:
+        assert self._executable is not None
+        # Under landlock we need this for execve to work.
+        # We use `self._executable` because it is copied when caching executors, but other properties are not.
+        return super().get_fs() + [ExactFile(self._executable)]
```
Xyene, line 94: The fact that both `get_executable` and `get_compiled_file` exist but return possibly different values is an abomination, and we should fix this. You don't have to do so in this PR, but please leave a comment to the effect of this being a hack to work around that.

dmoj/executors/base_executor.py
```diff
@@ -33,6 +33,7 @@
     ExactFile('/dev/urandom'),
     ExactFile('/dev/random'),
     *USR_DIR,
+    RecursiveDir('/bin'),  # required under landlock when /bin is not a symlink, since we check execve.
```
Xyene, line 36: Fine to not leave a comment here.

dmoj/cptbox/helper.cpp
```diff
@@ -90,13 +91,62 @@ int cptbox_child_run(const struct child_config *config) {
...
+        goto seccomp_setup;
+    }
+
+    ruleset_fd = landlock_create_ruleset(&ruleset_attr, sizeof(ruleset_attr), 0);
+    if (ruleset_fd < 0) {
+        perror("Failed to create a ruleset");
+        return PTBOX_SPAWN_FAIL_LANDLOCK;
+    }
+
+    // Note: WRITE must imply READ. This is required because even if one rule allows writing and another allows reading,
```
Xyene, line 115: The `O_TRUNC` handling in Landlock scares me. Could we add some testcases with a custom grader that make sure that we don't allow truncating etc.?

dmoj/cptbox/helper.cpp
Riolku: How? Consumers of this function throw a python error, but what would be appropriate to do here?

dmoj/cptbox/tracer.py
Riolku: Unsure. `security` defaults to None. I don't know why.

dmoj/executors/SCALA.py
Riolku: SCALA takes a different path under landlock last I checked, and that path requires `mincore`. I believe it had to do with how landlock deals with `/proc/<pid>`.

We decided at some point to simply allow the call.

dmoj/cptbox/helper.cpp
Riolku: I agree that this location is strange. Where should it live?

dmoj/executors/RUST.py
Riolku: The problem isn't because `get_executable` and `get_compiled_file` both exist. Indeed, I expect them to return the same file for RUST.

The reason they don't is because the cached executor isn't a full copy of the original. Instead, it simply sets `_executable` and `_dir` on the returned object, and nothing else.

For rust, that means that we try and lock a new directory, which is obviously wrong.

To bypass it I use `_executable` here, but the proper fix may be instead to look into the caching.

Here's the relevant line: https://github.com/DMOJ/judge-server/blob/master/dmoj/executors/compiled_executor.py#L58

Also, as for `get_executable` vs `get_compiled_file`, check out Java's usage of the methods:

https://github.com/DMOJ/judge-server/blob/master/dmoj/executors/java_executor.py#L82

Maybe these exist because Java's compiled file isn't the executable?

dmoj/cptbox/helper.cpp
Xyene: I think living in this file is probably fine, and it should return an enum. Then it should be consulted in the spawn routine alongside the Landlock version we have available. My concern here is more about subtly overloading the semantics of `get_landlock_version` than anything else.

We should bail if someone requests:

* seccomp or seccomp+landlock support on FreeBSD
* ptrace-only support on Linux
* seccomp+landlock on a Linux with too old a Landlock ABI

This sanity check could exist in `judge.py` alongside all the other sanity checks, but I don't feel strongly about that.

dmoj/cptbox/helper.cpp
Xyene: Got it. I think `has_landlock` should be implemented here rather than in Python, so we can return false order FREEBSD.

dmoj/cptbox/tracer.py
Xyene: My question was more: it seems weird that the default uninitialized state would be different from initializing with [].

dmoj/executors/SCALA.py
Xyene: Ack, thanks for explaining. Could we split this out in its own commit?

Xyene: Eh, I guess it kind of is already. Fine to leave it as-is.

dmoj/cptbox/tracer.py
Riolku: We could maybe avoid it by having a default of empty list in `_cptbox.pyx`, but I don't see how else to avoid it. `configure_files` is the function that sets the `landlock_*` attributes.

We could, in theory, have them defaulted to empty and then have this branch to set them to something, but I don't love that.

PR #908, opened 2021-09-12
cptbox: always set PR_SET_NO_NEW_PRIVS

Xyene:
We no longer support kernels older than 4.8 anyway. Also, the ifdef was
never being taken because we failed to include sys/prctl.h...


Xyene:
ok to test

quantum5 (approved):

PR #909, opened 2021-09-12
executors: disallow /dev/std{in,err,out}

Xyene:
These files are not POSIX, so runtimes should not be relying on their
presence.


kiritofeng:
ok to test

quantum5 (approved):
Seems legit. I think these were originally intended for symlinks or something to allow files to be aliased to the standard channels? That shouldn't be necessary now the we can convert open to dup if we really wanted to.

PR #917, opened 2021-09-15
fix java executors VM config detection

Riolku:
previously, an erroneous directory was chosen for the JVM config. To 
ensure a good path is chosen, we use a glob, and resolve the VM path.


kiritofeng:
ok to test

Xyene (approved):
LGTM, thanks

PR #918, opened 2021-09-15
executors: drop Groovy support

Xyene:
It has been broken for over a year now and never worked properly in the
first place, but nobody has complained.

Let's just remove it.


Xyene:
Apparently I unknowingly fixed it today...

PR #927, opened 2021-09-18
cptbox: make sys_utimensat emulation more accurate

quantum5:
This handles the case where times is NULL, which is valid and should not
generate EFAULT.


Xyene (approved):
LGTM

PR #928, opened 2021-09-18
Improve dmoj.utils.*

quantum5:
Most just adds `mypy` and switch to f-strings. Also start using `signal.strsignal` instead of `ctypes`.

### DIFF
--- a/dmoj/utils/os_ext.py
+++ b/dmoj/utils/os_ext.py
@@ -2,6 +2,7 @@ import ctypes
 import ctypes.util
 import os
 import signal
+from typing import Optional
 
 from dmoj.utils.unicode import utf8bytes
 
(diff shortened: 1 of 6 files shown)


dmoj/utils/os_ext.py
```diff
@@ -2,40 +2,48 @@
...
+from typing import Optional
 
 from dmoj.utils.unicode import utf8bytes
 
 OOM_SCORE_ADJ_MAX = 1000
 OOM_SCORE_ADJ_MIN = -1000
 
 
-def oom_score_adj(score, to=None):
+def oom_score_adj(score, to: Optional[int] = None) -> None:
```
kiritofeng, line 13: Should we annotate `score` as well?

kiritofeng (commented):
Otherwise LGTM

dmoj/utils/os_ext.py
quantum5: Yes.

kiritofeng (approved):
LGTM

PR #929, opened 2021-09-18
executors: implement sys_fchmodat sandboxing for compilers

quantum5:


Xyene (approved):
But leave the fixme comment, since it applies to the fchmod below.

PR #935, opened 2021-09-19
cptbox: drop support for FreeBSD < 8

quantum5:
The fcntl(F_CLOSEM) approach is only needed for FreeBSD releases out of
support for over 6 years, and the code path doesn't even work as it used
the wrong variable name.


Xyene (approved):
LGTM

PR #937, opened 2021-09-19
cptbox: ignore ESRCH from post_syscall handler

quantum5:
This deals with flakiness when a child process gets killed on Linux, which
manifests as randomly failing Go compiles. Since Go is not INTERCAL, we
strive to avoid such "random compiler bugs".


Xyene (approved):
LGTM

PR #939, opened 2021-09-19
cptbox: correctly handle sys_execve being allowed with seccomp

quantum5:
This avoids the hack of using a callback that returns True.


Xyene (approved):
LGTM

PR #940, opened 2021-09-19
executors: remove FIXME about posix_fadvise and posix_fallocate

quantum5:
* `posix_fadvise` is only a hint for caching, and does not modify the file.
* `posix_fallocate` fails if the fd is opened without write permission.


Xyene (approved):
LGTM

PR #941, opened 2021-09-19
cptbox: clang-format

quantum5:
Broken example: https://github.com/DMOJ/judge-server/runs/3671774719?check_suite_focus=true


kiritofeng (approved):
LGTM

PR #942, opened 2021-09-20
testsuite: reduce memory limit for C# MLE test

quantum5:
This should hopefully avoid flakiness with memory usage on FreeBSD.


Xyene (approved):
LGTM

PR #946, opened 2021-09-21
ci: fix ARM64 docker test

quantum5:
This PR also fixes all the issues that prevent executors from being used on arm64, as well as failing tests.


WallE256 (approved):
LGTM

PR #949, opened 2021-09-22
executors: use shutil.which instead of deprecated distutils

quantum5:
See [PEP 632](https://www.python.org/dev/peps/pep-0632/#migration-advice), which suggests replacing `distutils.spawn.find_executable` with `shutil.which`.


quantum5:
retest this please

Xyene (approved):
LGTM

PR #950, opened 2021-09-22
ci: remove useless functions from citest.py

quantum5:
Most notably, this removes a dependency on `distutils.version.LooseVersion`, which is deprecated in [PEP 632](https://www.python.org/dev/peps/pep-0632/).


Xyene (approved):
LGTM

PR #952, opened 2021-09-22
Declare types for base checker, problems, and graders

quantum5:


Riolku:
Maybe we want to bump this?

Xyene (approved):
This LGTM, I don't remember what the problem keeping this from being merged was anymore.

Issue #967, opened 2021-12-06
Auto-detect generator language when not explicitly specified


Xyene:
https://github.com/DMOJ/judge-server/blob/efc9f427e003f97c258a8aa82b6ce2d6de643ca5/dmoj/problem.py#L275

This would allow site-managed generators to be written in more than C++. We already store known extensions in the `Executor` classes.

PR #968, opened 2021-12-06
executors: add basic autodetection from filename

Xyene:
Ref https://github.com/DMOJ/judge-server/issues/967


dmoj/executors/__init__.py
```diff
@@ -15,6 +15,28 @@
...
 
+def by_ext(ext : str):
+    ext = ext.lower()
+
+    for name, executor in executors.items():
+        if name.lower() == ext:
+            return executor
+
+    for executor in executors.values():
+        if executor.ext == ext:
```
quantum5, line 26: Is this deterministic? I really don't want it to flip randomly between python 2 and python 3.

Xyene: Added a sort to be safe.

quantum5 (approved):

PR #982, opened 2022-01-04
Add executor for Lean 4

int-y1:
Coq can't do I/O. Agda requires a resource-intensive command `cabal install Agda`. Lean 3 can't generate an executable (but it can compile `.lean` to `.olean` for a performance boost). The next best language is Lean 4.

How to test this:
- Install Lean 4 to `/opt/lean`
  ```bash
  mkdir /opt/lean
  curl --compressed -L "https://github.com/leanprover/lean4/releases/download/v4.0.0-m2/lean-4.0.0-m2-linux.tar.gz" | tar xz -C /opt/lean --strip-components=1
  export PATH="/opt/lean/bin:$PATH"
  ```
- Modify `.docker.test.py` and `dmoj/citest.py` to just test `LEAN4`
- Capture
  (nit: the version is actually `4.0.0-m2`)

(Lean 4 has some support for siggrading, so it's possible to host a theorem proving contest)

### DIFF
--- a/.docker.test.py
+++ b/.docker.test.py
@@ -6,7 +6,7 @@ from dmoj.citest import ci_test
 from dmoj.executors import get_available
 
 arch = platform.machine()
-ALLOW_FAIL = {'GASARM', 'JAVA9', 'JAVA10', 'OBJC'}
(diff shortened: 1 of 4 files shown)


quantum5:
ok to test

quantum5 (approved):

PR #983, opened 2022-01-07
Update docker test

int-y1:
- Remove `JAVA9` and `JAVA10`, because they're not added to `failed_executors`.
- `LEAN4` should only fail for aarch64.


quantum5:
ok to test

quantum5 (approved):

PR #984, opened 2022-01-15
Use socket.sendall to send packets

hieplpvip:
Currently, `_send_packet` doesn't check if all data has been sent. This is problematic when the number of problems is large, so the packet can't be sent in one pass. This PR fixes that.

### DIFF
--- a/dmoj/packet.py
+++ b/dmoj/packet.py
@@ -110,7 +110,6 @@ class PacketManager:
 
         log.info('Starting handshake with: [%s]:%s', self.host, self.port)
         self.input = self.conn.makefile('rb')
-        self.output = self.conn.makefile('wb', 0)
         self.handshake(problems, versions, self.name, self.key)
         log.info('Judge "%s" online: [%s]:%s', self.name, self.host, self.port)
 
@@ -238,7 +237,8 @@ class PacketManager:
         raw = zlib.compress(utf8bytes(json.dumps(packet)))
         with self._lock:
             try:
-                self.output.writelines((PacketManager.SIZE_PACK.pack(len(raw)), raw))
+                assert self.conn is not None
+                self.conn.sendall(PacketManager.SIZE_PACK.pack(len(raw)) + raw)
             except Exception:  # connection reset by peer
                 log.exception('Exception while sending packet to site, will not attempt to reconnect! Quitting judge.')
                 os._exit(1)


dmoj/packet.py
```diff
@@ -238,7 +238,13 @@ def _send_packet(self, packet: dict):
         raw = zlib.compress(utf8bytes(json.dumps(packet)))
         with self._lock:
             try:
-                self.output.writelines((PacketManager.SIZE_PACK.pack(len(raw)), raw))
```
quantum5, line 241: Python is supposed to send everything with this call. What exactly is the problem you are seeing?

hieplpvip: We host a very large number of problems, so the handshake packet is quite large (over 100KBs). Everything is fine if both the judge and bridged are run on the same server, but when the judge is moved to a separate server, it fails to handshake.

From what I understand after reading CPython source code, `writelines` just calls `write` for every line, which in turn calls `socket.send`. The document for `socket.send` clearly notes that "Applications are responsible for checking that all data has been sent; if only some of the data was transmitted, the application needs to attempt delivery of the remaining data." ([ref](https://docs.python.org/3/library/socket.html#socket.socket.send)). In our case, `raw` is so large that `socket.send(raw)` could not send everything in one pass.

P/s: Before coming up with this fix, we have to monkey patch like this:

```
self.output.writelines((PacketManager.SIZE_PACK.pack(len(raw)), raw[:100000], raw[100000:]))
```

which handshakes successfully, so the problem seems to be indeed due to `socket.send` unable to send everything at once.

quantum5: This sounds like a Python bug with the makefile interface to sockets because it's supposed to make sockets look like normal streams and handle incomplete transmission. Out of curiosity, what version are you using and how many problems do you have?

hieplpvip: The judge uses Python 3.9.7. Currently, we have around 9000 problems. You can also reproduce this by adding an extremely long string to the packet:

```py
self._send_packet({'name': 'handshake', 'problems': problems, 'executors': runtimes, 'id': id, 'key': key, 'dummy': 'a' * 1000000000})
```

Btw, where do you get the info that "it's supposed to make sockets look like normal streams and handle incomplete transmission"? I can't seem to find it.

quantum5: Because `makefile` is supposed to return a file object and file objects do not have this weird behaviour in which a write could not complete fully. In this case what I would actually recommend is just getting rid of the file objects and use `socket.sendall`.

hieplpvip: Fixed

leduythuccs:
Could DMOJ please merge this PR, I'm having the same issue :'(

quantum5 (approved):

quantum5:
ok to test

quantum5:
Thanks for contributing to DMOJ!

PR #987, opened 2022-01-17
ci: improve automatic syscall PR generation

quantum5:
Basically don't say Linux in the PR title or commit message since we generate FreeBSD too now.


Xyene (approved):
LGTM

PR #989, opened 2022-01-30
Implement debug mode with debug logging level

quantum5:
This avoids spam but allows it when explicitly asked for.


Xyene (approved):
LGTM

PR #990, opened 2022-01-30
Switching to using memfd for input data

### LINKED ISSUE #835: Use memfd for file data
We can create a memfd pointing to decompressed problem data, and pass that off to the submissions.

Pros:

* no more piping necessary in judge; fewer context switches as a result
* submissions can seek the fd
* still no disk thrashing

Cons(?):

* 3.7+ kernel necessary
* extra memcpy

quantum5:
Breaking changes:

* All interactors using `_interact_with_process`

This closes #835.

### DIFF
--- a/dmoj/cptbox/helper.cpp
+++ b/dmoj/cptbox/helper.cpp
@@ -332,7 +332,7 @@ int memory_fd_create(void) {
 #ifdef __FreeBSD__
     char filename[] = "/tmp/cptbox-memoryfd-XXXXXXXX";
     int fd = mkstemp(filename);
-    if (fd > 0)
+    if (fd >= 0)
         unlink(filename);
     return fd;
 #else
--- a/dmoj/cptbox/helper.cpp
+++ b/dmoj/cptbox/helper.cpp
@@ -328,19 +328,16 @@ char *bsd_get_proc_fdno(pid_t pid, int fdno) {
     return bsd_get_proc_fd(pid, 0, fdno);
 }
 
-int memory_fd_create(void) {
+int cptbox_memfd_create(void) {
 #ifdef __FreeBSD__
-    char filename[] = "/tmp/cptbox-memoryfd-XXXXXXXX";
-    int fd = mkstemp(filename);
-    if (fd >= 0)
-        unlink(filename);
-    return fd;
+    errno = ENOSYS;
+    return -1;
 #else
     return memfd_create("cptbox memory_fd", MFD_ALLOW_SEALING);
 #endif
 }
 
-int memory_fd_seal(int fd) {
+int cptbox_memfd_seal(int fd) {
 #ifdef __FreeBSD__
     errno = ENOSYS;
     return -1;
--- a/dmoj/cptbox/utils.py
+++ b/dmoj/cptbox/utils.py
@@ -1,11 +1,42 @@
+import errno
 import io
+import mmap
+import os
+from typing import Optional
 
 from dmoj.cptbox._cptbox import memory_fd_create, memory_fd_seal
 
 
 class MemoryIO(io.FileIO):
-    def __init__(self) -> None:
+    def __init__(self, prefill: Optional[bytes] = None, seal=False) -> None:
         super().__init__(memory_fd_create(), 'r+')
+        if prefill:
+            self.write(prefill)
+        if seal:
+            self.seal()
 
     def seal(self) -> None:
-        memory_fd_seal(self.fileno())
+        fd = self.fileno()
+        try:
+            memory_fd_seal(fd)
+        except OSError as e:
+            if e.errno == errno.ENOSYS:
+                # FreeBSD
+                self.seek(0, os.SEEK_SET)
+                return
+            raise
+
+        new_fd = os.open(f'/proc/self/fd/{fd}', os.O_RDONLY)
+        try:
+            os.dup2(new_fd, fd)
+        finally:
+            os.close(new_fd)
+
+    def to_bytes(self) -> bytes:
+        try:
+            with mmap.mmap(self.fileno(), 0, access=mmap.ACCESS_READ) as f:
+                return bytes(f)
+        except ValueError as e:
+            if e.args[0] == 'cannot mmap an empty file':
+                return b''
+            raise
--- a/dmoj/cptbox/utils.py
+++ b/dmoj/cptbox/utils.py
@@ -32,6 +32,9 @@ class MemoryIO(io.FileIO):
         finally:
             os.close(new_fd)
 
+    def to_path(self) -> str:
+        return f'/proc/{os.getpid()}/fd/{self.fileno()}'
+
     def to_bytes(self) -> bytes:
         try:
             with mmap.mmap(self.fileno(), 0, access=mmap.ACCESS_READ) as f:
--- a/dmoj/cptbox/utils.py
+++ b/dmoj/cptbox/utils.py
@@ -1,39 +1,42 @@
-import errno
 import io
 import mmap
 import os
+from abc import ABCMeta, abstractmethod
+from tempfile import NamedTemporaryFile, TemporaryFile
 from typing import Optional
 
-from dmoj.cptbox._cptbox import memory_fd_create, memory_fd_seal
+from dmoj.cptbox._cptbox import memfd_create, memfd_seal
 
 
-class MemoryIO(io.FileIO):
-    def __init__(self, prefill: Optional[bytes] = None, seal=False) -> None:
-        super().__init__(memory_fd_create(), 'r+')
+def _make_fd_readonly(fd):
+    new_fd = os.open(f'/proc/self/fd/{fd}', os.O_RDONLY)
+    try:
+        os.dup2(new_fd, fd)
+    finally:
+        os.close(new_fd)
+
+
+class MmapableIO(io.FileIO, metaclass=ABCMeta):
+    def __init__(self, fd, *, prefill: Optional[bytes] = None, seal=False) -> None:
+        super().__init__(fd, 'r+')
+
         if prefill:
             self.write(prefill)
         if seal:
             self.seal()
 
+    @classmethod
+    @abstractmethod
+    def usable_with_name(cls) -> bool:
+        ...
+
+    @abstractmethod
     def seal(self) -> None:
-        fd = self.fileno()
-        try:
-            memory_fd_seal(fd)
-        except OSError as e:
-            if e.errno == errno.ENOSYS:
-                # FreeBSD
-                self.seek(0, os.SEEK_SET)
-                return
-            raise
+        ...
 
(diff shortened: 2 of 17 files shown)


dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
     def seal(self) -> None:
-        memory_fd_seal(self.fileno())
+        fd = self.fileno()
+        try:
+            memory_fd_seal(fd)
+        except OSError as e:
+            if e.errno == errno.ENOSYS:
+                # FreeBSD
+                self.seek(0, os.SEEK_SET)
+                return
```
Riolku, line 36: I'm not quite sure what this does. Does it deserve more of a comment?

dmoj/problem.py
```diff
@@ -330,17 +346,33 @@ def _run_generator(self, gen, args=None):
...
 
     def input_data(self):
+        return self.input_data_fd().to_bytes()
+
+    def input_data_fd(self):
+        if not self._input_data_fd:
+            self._input_data_fd = self._make_input_data_fd()
+        return self._input_data_fd
+
+    def _make_input_data_fd(self):
```
Riolku, line 356: Maybe this needs to be typed to return `MemoryIO` now?

dmoj/cptbox/helper.cpp
```diff
@@ -306,7 +306,7 @@ int memory_fd_create(void) {
 #ifdef __FreeBSD__
```
Riolku, line 306: Actually, is this function called on FreeBSD anymore? Are you creating the tempfile in Python instead?

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
+            raise
+
+        new_fd = os.open(f'/proc/self/fd/{fd}', os.O_RDONLY)
+        try:
+            os.dup2(new_fd, fd)
+        finally:
+            os.close(new_fd)
```
Riolku, line 43: Maybe we could use a comment about why this dup is needed. Also, why isn't it implemented in the C function?

dmoj/executors/shell_executor.py
```diff
@@ -21,10 +21,10 @@ def get_fs(self):
     def get_allowed_syscalls(self):
         return super().get_allowed_syscalls() + ['fork', 'waitpid', 'wait4']
 
-    def get_security(self, launch_kwargs=None):
+    def get_security(self, launch_kwargs=None, extra_fs=None):
```
Riolku, line 24: Not that this is a direct result of this PR, but maybe this should be converted to `**kwargs` instead?

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
-    def __init__(self) -> None:
-        super().__init__(memory_fd_create(), 'r+')
+    _name: Optional[str] = None
+
+    def __init__(self, prefill: Optional[bytes] = None, seal=False) -> None:
+        if FREEBSD:
+            with NamedTemporaryFile(delete=False) as f:
+                self._name = f.name
+                super().__init__(os.dup(f.fileno()), 'r+')
+        else:
```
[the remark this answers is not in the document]
quantum5: Definitely. Otherwise the fd gets closed or the file gets unlinked, respectively.

dmoj/cptbox/utils.py
quantum5: C code is just that much more painful to maintain.

dmoj/problem.py
quantum5: Any file object will do, `MemoryIO` is just a special hack.

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
+from dmoj.cptbox.tracer import FREEBSD
 
 
 class MemoryIO(io.FileIO):
-    def __init__(self) -> None:
-        super().__init__(memory_fd_create(), 'r+')
+    _name: Optional[str] = None
+
+    def __init__(self, prefill: Optional[bytes] = None, seal=False) -> None:
+        if FREEBSD:
```
Xyene, line 16: I think we either want to branch on some sort of "expected size" here, or figure out something to make the memory backing this file be more likely to be swapped out.

If the input is 5 GiB large, we don't want to buffer all of it in memory.

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
+import mmap
+import os
+from tempfile import NamedTemporaryFile
+from typing import Optional
 
 from dmoj.cptbox._cptbox import memory_fd_create, memory_fd_seal
+from dmoj.cptbox.tracer import FREEBSD
 
 
 class MemoryIO(io.FileIO):
```
Xyene, line 12: Where/how do we end up deleting the memory backing these objects once they're no longer necessary to grade a case?

quantum5: It's closed when `MemoryIO.close` is called or it's garbage collected, since `FileIO` constructor takes ownership of the file descriptor unless `closefd=False` is passed.

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
+        new_fd = os.open(f'/proc/self/fd/{fd}', os.O_RDONLY)
+        try:
+            os.dup2(new_fd, fd)
+        finally:
+            os.close(new_fd)
+
+    def close(self) -> None:
+        super().close()
+        if self._name:
+            os.unlink(self._name)
```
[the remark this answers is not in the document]
quantum5: The resulting file can't be reopened from another process on FreeBSD. Made this more explicit in the new version.

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
-        super().__init__(memory_fd_create(), 'r+')
+    _name: Optional[str] = None
+
+    def __init__(self, prefill: Optional[bytes] = None, seal=False) -> None:
+        if FREEBSD:
+            with NamedTemporaryFile(delete=False) as f:
+                self._name = f.name
+                super().__init__(os.dup(f.fileno()), 'r+')
+        else:
+            super().__init__(memory_fd_create(), 'r+')
```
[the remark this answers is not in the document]
quantum5: Changed to have both `NamedFileIO` and `MemoryIO`, with the latter being an alias of the former on FreeBSD.

dmoj/cptbox/utils.py
quantum5: Leaving this up to the caller for now. If the caller desires, it can use `NamedFileIO`.

dmoj/executors/shell_executor.py
quantum5: We can do that later.

dmoj/cptbox/utils.py
quantum5: Factored it to a function with an obvious name.

dmoj/cptbox/helper.cpp
quantum5: Let's keep this around for now, since I'd rather this function work on all platforms, as the detection logic for the FreeBSD case is now different. If FreeBSD implements `/proc/[pid]/fd` some day, this will magically work.

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,62 @@
...
 from dmoj.cptbox._cptbox import memory_fd_create, memory_fd_seal
+from dmoj.cptbox.tracer import FREEBSD
 
 
 class MemoryIO(io.FileIO):
-    def __init__(self) -> None:
-        super().__init__(memory_fd_create(), 'r+')
+    _name: Optional[str] = None
+
+    def __init__(self, prefill: Optional[bytes] = None, seal=False) -> None:
```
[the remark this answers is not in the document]
quantum5: Made keyword arguments required.

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,85 @@
...
+from tempfile import NamedTemporaryFile
+from typing import Optional
 
 from dmoj.cptbox._cptbox import memory_fd_create, memory_fd_seal
 
 
-class MemoryIO(io.FileIO):
-    def __init__(self) -> None:
-        super().__init__(memory_fd_create(), 'r+')
+def _make_fd_readonly(fd):
```
Xyene, line 12: This can fail silently.

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,85 @@
...
+        super().__init__(fd, 'r+')
+
+        if prefill:
+            self.write(prefill)
+        if seal:
+            self.seal()
+
+    def to_bytes(self) -> bytes:
+        try:
+            with mmap.mmap(self.fileno(), 0, access=mmap.ACCESS_READ) as f:
```
Xyene, line 35: How often do we expect this will be called? Should we `madvise(..., MADV_SEQUENTIAL)` here?

dmoj/cptbox/utils.py
```diff
@@ -1,11 +1,89 @@
...
+        with NamedTemporaryFile(delete=False) as f:
+            self._name = f.name
+            super().__init__(os.dup(f.fileno()), prefill=prefill, seal=seal)
+
+    def seal(self) -> None:
+        self.seek(0, os.SEEK_SET)
+
+    def close(self) -> None:
+        super().close()
+        os.unlink(self._name)
```
Xyene, line 56: Probably we should also implement `__del__`, since I'm not convinced we call `free_data` in all cases.

dmoj/cptbox/utils.py
quantum5: What do you mean? If `os.open` or `os.dup2` fails, `OSError` is raised.

dmoj/cptbox/utils.py
quantum5: It's implemented by default:

```pycon
Python 3.11.2 (main, Sep 14 2024, 03:00:30) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from dmoj.cptbox.utils import NamedFileIO
>>> NamedFileIO()
<_io.FileIO name=4 mode='rb+' closefd=True>
>>> 0
unlinking
0
```

dmoj/cptbox/utils.py
quantum5: Not very often, it's mostly for compatibility with old checkers etc.

Xyene: Sounds like "very often" to me, but happy to punt on this. I worry we'll hit issues with gigabyte-sized generator inputs that also have checkers, since this doubles the memory requirement.

quantum5: It's less of a problem than it looks. In the standard grader, we pass this magic to checkers: `judge_input=LazyBytes(case.input_data)`. We only pay for this if the checker actually reads `judge_input`.

Xyene (approved):

PR #992, opened 2022-02-10
cptbox: implement setting child affinity

Xyene:


quantum5 (approved):
Whatever, 64 cores is enough for the next little while...

PR #994, opened 2022-02-13
cptbox: track and report context switches

Xyene:
image


quantum5 (approved):
LGTM, should probably deduct the switches pre-exec but that can be done later.

PR #996, opened 2022-02-15
Isolate refactor

Riolku:
This PR refactors isolate.py, including moving the recently introduced FilesystemSyscallKind into its own file, moving the compiler isolate tracer into cptbox, and splitting up the logic overall.

The main hope is that this PR will make coding on top of this system easier, especially in conjunction with the landlock PR, for implementation of a debug mode.

### DIFF
--- a/dmoj/cptbox/compiler_isolate.py
+++ b/dmoj/cptbox/compiler_isolate.py
@@ -1,0 +1,149 @@
+import struct
+import sys
+
+from dmoj.cptbox._cptbox import AT_FDCWD, Debugger
+from dmoj.cptbox.filesystem_policies import ExactFile, FilesystemPolicy, RecursiveDir
+from dmoj.cptbox.handlers import ACCESS_EFAULT, ACCESS_EPERM, ALLOW
+from dmoj.cptbox.isolate import DeniedSyscall, FilesystemSyscallKind, IsolateTracer
+from dmoj.cptbox.syscalls import *
+from dmoj.cptbox.tracer import AdvancedDebugger
+from dmoj.executors.base_executor import BASE_FILESYSTEM, BASE_WRITE_FILESYSTEM
+
+
+UTIME_OMIT = (1 << 30) - 2
+
+
+class CompilerIsolateTracer(IsolateTracer):
+    def __init__(self, tmpdir, read_fs, write_fs, *args, **kwargs):
+        read_fs += BASE_FILESYSTEM + [
+            RecursiveDir(tmpdir),
+            ExactFile('/bin/strip'),
+            RecursiveDir('/usr/x86_64-linux-gnu'),
+        ]
+        write_fs += BASE_WRITE_FILESYSTEM + [RecursiveDir(tmpdir)]
+        super().__init__(read_fs, *args, write_fs=write_fs, **kwargs)
+
+        self.update(
+            {
+                # Process spawning system calls
+                sys_fork: ALLOW,
+                sys_vfork: ALLOW,
+                sys_execve: ALLOW,
+                sys_getcpu: ALLOW,
+                sys_getpgid: ALLOW,
+                # Directory system calls
+                sys_mkdir: self.handle_file_access(FilesystemSyscallKind.WRITE, file_reg=0),
+                sys_mkdirat: self.handle_file_access_at(FilesystemSyscallKind.WRITE, dir_reg=0, file_reg=1),
+                sys_rmdir: self.handle_file_access(FilesystemSyscallKind.WRITE, file_reg=0),
+                # Linking system calls
+                sys_link: self.handle_file_access(FilesystemSyscallKind.WRITE, file_reg=1),
+                sys_linkat: self.handle_file_access_at(FilesystemSyscallKind.WRITE, dir_reg=2, file_reg=3),
+                sys_unlink: self.handle_file_access(FilesystemSyscallKind.WRITE, file_reg=0),
+                sys_unlinkat: self.handle_file_access_at(FilesystemSyscallKind.WRITE, dir_reg=0, file_reg=1),
+                sys_symlink: self.handle_file_access(FilesystemSyscallKind.WRITE, file_reg=1),
+                # Miscellaneous other filesystem system calls
+                sys_chdir: self.handle_file_access(FilesystemSyscallKind.READ, file_reg=0),
+                sys_chmod: self.handle_file_access(FilesystemSyscallKind.WRITE, file_reg=0),
+                sys_utimensat: self.do_utimensat,
+                sys_umask: ALLOW,
+                sys_flock: ALLOW,
+                sys_fsync: ALLOW,
+                sys_fadvise64: ALLOW,
+                sys_fchmodat: self.handle_file_access_at(FilesystemSyscallKind.WRITE, dir_reg=0, file_reg=1),
+                sys_fchmod: self.handle_fchmod,
+                sys_fallocate: ALLOW,
+                sys_ftruncate: ALLOW,
+                sys_rename: self.handle_rename,
+                sys_renameat: self.handle_renameat,
+                # I/O system calls
+                sys_readv: ALLOW,
+                sys_pwrite64: ALLOW,
+                sys_sendfile: ALLOW,
+                # Event loop system calls
+                sys_epoll_create: ALLOW,
+                sys_epoll_create1: ALLOW,
+                sys_epoll_ctl: ALLOW,
+                sys_epoll_wait: ALLOW,
+                sys_epoll_pwait: ALLOW,
+                sys_timerfd_settime: ALLOW,
+                sys_eventfd2: ALLOW,
+                sys_waitid: ALLOW,
+                sys_wait4: ALLOW,
+                # Network system calls, we don't sandbox these
+                sys_socket: ALLOW,
+                sys_socketpair: ALLOW,
+                sys_connect: ALLOW,
+                sys_setsockopt: ALLOW,
+                sys_getsockname: ALLOW,
+                sys_sendmmsg: ALLOW,
+                sys_recvfrom: ALLOW,
+                sys_sendto: ALLOW,
+                # Miscellaneous other system calls
+                sys_msync: ALLOW,
+                sys_clock_nanosleep: ALLOW,
+                sys_memfd_create: ALLOW,
+                sys_rt_sigsuspend: ALLOW,
+            }
+        )
+
+        # FreeBSD-specific syscalls
+        if 'freebsd' in sys.platform:
+            self.update(
+                {
+                    sys_rfork: ALLOW,
+                    sys_procctl: ALLOW,
+                    sys_cap_rights_limit: ALLOW,
+                    sys_posix_fadvise: ALLOW,
+                    sys_posix_fallocate: ALLOW,
+                    sys_setrlimit: ALLOW,
+                    sys_cap_ioctls_limit: ALLOW,
+                    sys_cap_fcntls_limit: ALLOW,
+                    sys_cap_enter: ALLOW,
+                    sys_utimes: self.handle_file_access(FilesystemSyscallKind.WRITE, file_reg=0),
+                }
+            )
+
+    def handle_rename(self, debugger: Debugger) -> None:
+        self.access_check(self._write_fs_jail_getter, self._dirfd_getter_cwd, file_reg=0)(debugger)
+        self.access_check(self._write_fs_jail_getter, self._dirfd_getter_cwd, file_reg=1)(debugger)
+
+    def handle_renameat(self, debugger: Debugger) -> None:
+        self.access_check(self._write_fs_jail_getter, self._dirfd_getter_from_reg(0), file_reg=1)(debugger)
+        self.access_check(self._write_fs_jail_getter, self._dirfd_getter_from_reg(2), file_reg=3)(debugger)
+
+    def _write_fs_jail_getter(self, debugger: Debugger) -> FilesystemPolicy:
+        return self.write_fs_jail
+
+    def do_utimensat(self, debugger: AdvancedDebugger) -> None:
+        timespec = struct.Struct({32: '=ii', 64: '=QQ'}[debugger.address_bits])
+
+        # Emulate https://github.com/torvalds/linux/blob/v5.14/fs/utimes.c#L152-L161
+        times_ptr = debugger.uarg2
+        if times_ptr:
+            try:
+                buffer = debugger.readbytes(times_ptr, timespec.size * 2)
+            except OSError:
+                raise DeniedSyscall(ACCESS_EFAULT, f'Cannot read from times_ptr: {times_ptr}')
+
+            times = list(timespec.iter_unpack(buffer))
+            if times[0][1] == UTIME_OMIT and times[1][1] == UTIME_OMIT:
+                debugger.syscall = -1
+
+                def on_return():
+                    debugger.result = 0
+
+                debugger.on_return(on_return)
+
+        else:
+            # Emulate https://github.com/torvalds/linux/blob/v5.14/fs/utimes.c#L142-L143
+            if debugger.uarg0 != AT_FDCWD and not debugger.uarg1:
+                path = self._getfd_pid(debugger.tid, debugger.uarg0)
+                if not self.write_fs_jail.check(path):
+                    raise DeniedSyscall(ACCESS_EPERM, f'Denying access to {path}')
+            else:
+                return self.handle_file_access_at(FilesystemSyscallKind.WRITE, dir_reg=0, file_reg=1)(debugger)
+
+    def handle_fchmod(self, debugger: Debugger) -> None:
+        path = self._getfd_pid(debugger.tid, debugger.uarg0)
+        if not self.write_fs_jail.check(path):
+            raise DeniedSyscall(ACCESS_EPERM, f'Denying access to {path}')
--- a/dmoj/cptbox/isolate.py
+++ b/dmoj/cptbox/isolate.py
@@ -2,15 +2,13 @@ import logging
 import os
 import sys
 from enum import Enum
-from operator import attrgetter
-from typing import Optional, Tuple
+from typing import Any, Callable, Mapping
 
 from dmoj.cptbox._cptbox import AT_FDCWD, Debugger, bsd_get_proc_cwd, bsd_get_proc_fdno
 from dmoj.cptbox.filesystem_policies import FilesystemPolicy
 from dmoj.cptbox.handlers import (
     ACCESS_EACCES,
     ACCESS_EFAULT,
-    ACCESS_EINVAL,
     ACCESS_ENAMETOOLONG,
     ACCESS_ENOENT,
     ACCESS_EPERM,
@@ -18,6 +16,7 @@ from dmoj.cptbox.handlers import (
     ErrnoHandlerCallback,
 )
 from dmoj.cptbox.syscalls import *
+from dmoj.cptbox.syscalls import by_id
 from dmoj.cptbox.tracer import HandlerCallback, MaxLengthExceeded
 from dmoj.utils.unicode import utf8text
 
@@ -34,8 +33,12 @@ except AttributeError:
 class FilesystemSyscallKind(Enum):
     READ = 1
     WRITE = 2
-    OPEN = 3
-    FSTAT = 4
+
+
+AccessChecker = Callable[[Debugger], None]
+
+FSJailGetter = Callable[[Debugger], FilesystemPolicy]
+DirFDGetter = Callable[[Debugger], int]
 
 
 class IsolateTracer(dict):
@@ -56,22 +59,22 @@ class IsolateTracer(dict):
         self.update(
             {
                 # Deny with report
-                sys_openat: self.check_file_access_at('openat', FilesystemSyscallKind.OPEN),
-                sys_open: self.check_file_access('open', 0, FilesystemSyscallKind.OPEN),
-                sys_faccessat: self.check_file_access_at('faccessat', FilesystemSyscallKind.READ),
-                sys_faccessat2: self.check_file_access_at('faccessat2', FilesystemSyscallKind.READ),
-                sys_access: self.check_file_access('access', 0, FilesystemSyscallKind.READ),
-                sys_readlink: self.check_file_access('readlink', 0, FilesystemSyscallKind.READ),
-                sys_readlinkat: self.check_file_access_at('readlinkat', FilesystemSyscallKind.READ),
-                sys_stat: self.check_file_access('stat', 0, FilesystemSyscallKind.READ),
-                sys_stat64: self.check_file_access('stat64', 0, FilesystemSyscallKind.READ),
-                sys_lstat: self.check_file_access('lstat', 0, FilesystemSyscallKind.READ),
-                sys_lstat64: self.check_file_access('lstat64', 0, FilesystemSyscallKind.READ),
-                sys_fstatat: self.check_file_access_at('fstatat', FilesystemSyscallKind.FSTAT),
-                sys_statx: self.check_file_access_at('statx', FilesystemSyscallKind.FSTAT),
-                sys_tgkill: self.do_kill,
-                sys_kill: self.do_kill,
-                sys_prctl: self.do_prctl,
+                sys_openat: self.handle_openat(dir_reg=0, file_reg=1, flag_reg=2),
+                sys_open: self.handle_open(file_reg=0, flag_reg=1),
+                sys_faccessat: self.handle_file_access_at(FilesystemSyscallKind.READ, dir_reg=0, file_reg=1),
+                sys_faccessat2: self.handle_file_access_at(FilesystemSyscallKind.READ, dir_reg=0, file_reg=1),
+                sys_access: self.handle_file_access(FilesystemSyscallKind.READ, file_reg=0),
+                sys_readlink: self.handle_file_access(FilesystemSyscallKind.READ, file_reg=0),
+                sys_readlinkat: self.handle_file_access_at(FilesystemSyscallKind.READ, dir_reg=0, file_reg=1),
+                sys_stat: self.handle_file_access(FilesystemSyscallKind.READ, file_reg=0),
+                sys_stat64: self.handle_file_access(FilesystemSyscallKind.READ, file_reg=0),
+                sys_lstat: self.handle_file_access(FilesystemSyscallKind.READ, file_reg=0),
+                sys_lstat64: self.handle_file_access(FilesystemSyscallKind.READ, file_reg=0),
+                sys_fstatat: self.handle_fstat(dir_reg=0, file_reg=1),
+                sys_statx: self.handle_fstat(dir_reg=0, file_reg=1),
+                sys_tgkill: self.handle_kill,
+                sys_kill: self.handle_kill,
+                sys_prctl: self.handle_prctl,
                 sys_read: ALLOW,
                 sys_pread64: ALLOW,
                 sys_write: ALLOW,
@@ -159,7 +162,7 @@ class IsolateTracer(dict):
                 sys_llseek: ALLOW,
                 sys_fcntl64: ALLOW,
                 sys_time: ALLOW,
-                sys_prlimit64: self.do_prlimit,
+                sys_prlimit64: self.handle_prlimit,
                 sys_getdents64: ALLOW,
             }
         )
@@ -180,8 +183,8 @@ class IsolateTracer(dict):
                     sys_setcontext: ALLOW,
                     sys_pread: ALLOW,
                     sys_fsync: ALLOW,
-                    sys_shm_open: self.check_file_access('shm_open', 0, FilesystemSyscallKind.OPEN),
-                    sys_shm_open2: self.check_file_access('shm_open2', 0, FilesystemSyscallKind.OPEN),
+                    sys_shm_open: self.handle_open(file_reg=0, flag_reg=1),
+                    sys_shm_open2: self.handle_open(file_reg=0, flag_reg=1),
                     sys_cpuset_getaffinity: ALLOW,
                     sys_thr_new: ALLOW,
                     sys_thr_exit: ALLOW,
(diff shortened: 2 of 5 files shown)


quantum5:
ok to test

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
-
-        def check(debugger: Debugger) -> bool:
-            file, error = self.read_path(syscall, debugger, getattr(debugger, 'uarg%d' % argument))
-            if error is not None:
-                return error
-
-            file, error = self._file_access_check(file, debugger, fs_jail(debugger))
-            if not error:
-                return True
+    def handle_file_access(self, syscall: str, kind: FilesystemSyscallKind, file_reg: int = 0) -> HandlerCallback:
```
Xyene, line 212:
```suggestion
    def handle_file_access(self, syscall: str, *, kind: FilesystemSyscallKind, file_reg: int = 0) -> HandlerCallback:
```

and fix resulting errors

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+        return self.wrap_access_check(
+            syscall,
+            self.main_access_check(syscall, self._fs_jail_getter_from_kind(kind), self._dirfd_getter_cwd, file_reg),
+        )
 
-            log.debug('Denied access via syscall %s (error: %s): %s', syscall, error.error_name, file)
-            return error(debugger)
+    def handle_file_access_at(
+        self, syscall: str, kind: FilesystemSyscallKind, dir_reg: int = 0, file_reg: int = 1
```
Xyene, line 219:
```suggestion
        self, syscall: str, *, kind: FilesystemSyscallKind, dir_reg: int, file_reg: int
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+    ) -> HandlerCallback:
+        return self.wrap_access_check(
+            syscall,
+            self.main_access_check(
+                syscall, self._fs_jail_getter_from_kind(kind), self._dirfd_getter_from_reg(dir_reg), file_reg
+            ),
+        )
 
-        return check
+    def handle_open(self, syscall: str, file_reg: int = 0, flag_reg: int = 1) -> HandlerCallback:
```
Xyene, line 228:
```suggestion
    def handle_open(self, syscall: str, *, file_reg: int, flag_reg: int) -> HandlerCallback:
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+        return self.wrap_access_check(
+            syscall,
+            self.main_access_check(
+                syscall, self._fs_jail_getter_from_open_flags_reg(flag_reg), self._dirfd_getter_cwd, file_reg
+            ),
+        )
 
-    def check_file_access_at(self, syscall, kind, file_reg=1) -> HandlerCallback:
-        fs_jail = self._fs_jail(attrgetter('uarg2'), kind)
+    def handle_openat(self, syscall: str, dir_reg: int = 0, file_reg: int = 1, flag_reg: int = 2) -> HandlerCallback:
```
Xyene, line 236:
```suggestion
    def handle_openat(self, syscall: str, *, dir_reg: int, file_reg: int, flag_reg: int) -> HandlerCallback:
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
             # we otherwise wouldn't have.
-            if kind == FilesystemSyscallKind.FSTAT and file == '' and debugger.uarg3 & AT_EMPTY_PATH:
+            if rel_file == '' and debugger.uarg3 & AT_EMPTY_PATH:
                 # If pathname is an empty string, operate on the file referred to
                 # by dirfd (which may have been obtained using the open(2) O_PATH
                 # flag). In this case, dirfd can refer to any type of file, not
                 # just a directory, and the behavior of fstatat() is similar to
                 # that of fstat(). If dirfd is AT_FDCWD, the call operates on the
                 # current working directory.
+                # We already allowed this one way or another, don't check again
```
Xyene, line 265:
```suggestion
                # We already allowed this one way or another, don't check again.
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+    def _fs_jail_from_open_flags(self, open_flags: int) -> FilesystemPolicy:
+        for flag in open_write_flags:
+            # Strict equality is necessary here, since e.g. O_TMPFILE has multiple bits set,
+            # and O_DIRECTORY & O_TMPFILE > 0.
+            if open_flags & flag == flag:
+                return self.write_fs_jail
+
+        return self.read_fs_jail
+
+    def _fs_jail_from_kind(self, kind: FilesystemSyscallKind) -> FilesystemPolicy:
```
Xyene, line 330: This function is used in just one place, inline it

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
-        projected = normalized = '/' + os.path.normpath(file).lstrip('/')
-        if normalized.startswith('/proc/self'):
-            file = os.path.join(f'/proc/{debugger.tid}', os.path.relpath(file, '/proc/self'))
-            projected = '/' + os.path.normpath(file).lstrip('/')
-        elif normalized.startswith(f'/proc/{debugger.tid}/'):
-            # If the child process uses /proc/getpid()/foo, set the normalized path to be /proc/self/foo.
-            # Access rules can more easily check /proc/self.
-            normalized = os.path.join('/proc/self', os.path.relpath(file, f'/proc/{debugger.tid}'))
-        real = os.path.realpath(file)
+        # We need to project /proc/self.
```
Xyene, line 353: It's worth keeping the original comment here

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
         PR_GET_DUMPABLE = 3
         PR_SET_NAME = 15
         PR_GET_NAME = 16
         PR_SET_THP_DISABLE = 41
         PR_SET_VMA = 0x53564D41  # Used on Android
         return debugger.arg0 in (PR_GET_DUMPABLE, PR_SET_NAME, PR_GET_NAME, PR_SET_THP_DISABLE, PR_SET_VMA)
+
+
+class AccessFailure(Exception):
+    def __init__(self, handler: Callable, severity, reason_format: str, *reason_args):
```
Xyene, line 481: The severity is always warning, so no need to have it as a parameter.

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+            # in cptbox.  Either way this call should not be allowed; if the path
+            # was indeed NULL we can end the request before it gets to the kernel
+            # without any downside, and if it was *not* NULL and we failed to read
+            # it, then we should *definitely* stop the call here.
+            if file is None:
+                raise AccessFailure(ACCESS_EFAULT, logging.WARNING, 'Unreadable or NULL path')
+
+            return file
+
+    def get_dir(self, debugger: Debugger, dirfd: int) -> str:
```
Xyene, line 445:
```suggestion
    def get_dir(self, debugger: Debugger, *, dirfd: int) -> str:
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+        return self.rewrite_path_prefix(path, f'/proc/{debugger.tid}', '/proc/self')
+
+    # if path starts with prefix_src, change it to the equivalent path starting with prefix_dst instead
+    def rewrite_path_prefix(self, path: str, prefix_src: str, prefix_dst: str) -> str:
+        if path.startswith(prefix_src):
+            return os.path.join(prefix_dst, os.path.relpath(path, prefix_src))
+        else:
+            return path
+
+    def get_rel_file(self, debugger: Debugger, reg: int) -> str:
```
Xyene, line 416:
```suggestion
    def get_rel_file(self, debugger: Debugger, *, reg: int) -> str:
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+        return True
 
-            if not fs_jail.check(real):
-                return real, ACCESS_EACCES
+    def log_disallowed_file(self, syscall: str, file: str, checked_file: str) -> None:
+        log.debug('Denied access via %s to: %s, actually checked: %s', syscall, file, checked_file)
 
-        return normalized, None
+    def get_reg_value(self, debugger: Debugger, reg: int) -> int:
+        return getattr(debugger, 'uarg%d' % reg)
```
Xyene, line 399: Please inline this, or if it really hurts your eyes create a `Debugger.uarg` that implements `[x]` as `getattr(self, 'uarg%d' % reg)`.

dmoj/cptbox/isolate.py
Xyene: I'd appreciate leaving this code as it was, it's harder to read now that one has to keep going to read short functions (that aren't reused elsewhere for it to be worth it)

Xyene: (it's not obvious from the function names the nuance of what they actually do, and the nuance is complicated enough that encoding it in a name seems impossible)

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+        return self.read_fs_jail
+
+    def _fs_jail_from_kind(self, kind: FilesystemSyscallKind) -> FilesystemPolicy:
+        if kind == FilesystemSyscallKind.READ:
+            return self.read_fs_jail
+        else:
+            assert kind == FilesystemSyscallKind.WRITE, 'Bad FilesystemSyscallKind'
+            return self.write_fs_jail
+
+    def main_access_check(
```
Xyene, line 337: "main", as opposed to...? This is the only "public" access check function, feel free to call it `access_check` and the internal one just `_access_check`.

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
-            if not error:
-                return True
+            dirfd = self.get_reg_value(debugger, dir_reg)
+            full_path = self.get_full_path(debugger, rel_file, dirfd)
+            return self._raw_access_check(syscall, debugger, full_path, self.read_fs_jail)
+
+        return check
+
+    @staticmethod
+    def wrap_access_check_dec(syscall: str) -> Callable[[AccessChecker], HandlerCallback]:
```
Xyene, line 275: I'm not sure what `check_dec` is supposed to mean, I've read the code & tried to think, but failed

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+            syscall,
+            self.main_access_check(
+                syscall,
+                self._fs_jail_getter_from_open_flags_reg(flag_reg),
+                self._dirfd_getter_from_reg(dir_reg),
+                file_reg,
+            ),
+        )
 
+    def handle_fstat(self, syscall: str, dir_reg: int = 0, file_reg: int = 1) -> HandlerCallback:
```
Xyene, line 247:
```suggestion
    def handle_fstat(self, syscall: str, *, dir_reg: int = 0, file_reg: int = 1) -> HandlerCallback:
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+
+    @staticmethod
+    def wrap_access_check(syscall: str, access_check: AccessChecker) -> HandlerCallback:
+        def inner(debugger) -> bool:
+            try:
+                allowed = access_check(debugger)
+            except AccessFailure as failure:
+                failure.log_failure(syscall)
+                return failure.handle(debugger)
+            else:
```
Xyene, line 289: Doesn't need to be in an `else`, since the `except` returns.

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+                if not allowed:
+                    return ACCESS_EACCES(debugger)
+                else:
+                    return True
```
Xyene, line 293: Doesn't need to be in an `else`, since the `if` returns.

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+            return self.get_reg_value(debugger, reg)
+
+        return getter
 
-            log.debug('Denied access via syscall %s (error: %s): %s', syscall, error.error_name, file)
-            return error(debugger)
+    def _dirfd_getter_cwd(self, debugger: Debugger) -> int:
+        return AT_FDCWD
+
+    FSJailGetter = Callable[[Debugger], FilesystemPolicy]
```
Xyene, line 308: Hoist type declarations to the top of the file

dmoj/cptbox/isolate.py
```diff
@@ -34,8 +31,12 @@
 class FilesystemSyscallKind(Enum):
     READ = 1
     WRITE = 2
-    OPEN = 3
-    FSTAT = 4
+
+
+# Distinguished in typing from HandlerCallback
+# Denotes the difference between a access checker, which doesn't catch failures
+# And a callback, which must
```
Xyene, line 38: pls punctuate comments

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+    def _dirfd_getter_from_reg(self, reg: int) -> DirFDGetter:
+        def getter(debugger: Debugger) -> int:
+            return self.get_reg_value(debugger, reg)
+
+        return getter
 
-            log.debug('Denied access via syscall %s (error: %s): %s', syscall, error.error_name, file)
-            return error(debugger)
+    def _dirfd_getter_cwd(self, debugger: Debugger) -> int:
+        return AT_FDCWD
```
Xyene, line 306: How does this typecheck, returning an `int` instead of a `DirFDGetter`?

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
 
-        if normalized != real:
-            proc_dir = f'/proc/{debugger.tid}'
-            if real.startswith(proc_dir):
-                real = os.path.join('/proc/self', os.path.relpath(real, proc_dir))
+        return True
 
-            if not fs_jail.check(real):
-                return real, ACCESS_EACCES
+    def log_disallowed_file(self, syscall: str, file: str, checked_file: str) -> None:
```
Xyene, line 395: Used in only one place, please inline

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+
+    def _fs_jail_from_kind(self, kind: FilesystemSyscallKind) -> FilesystemPolicy:
+        if kind == FilesystemSyscallKind.READ:
+            return self.read_fs_jail
+        else:
+            assert kind == FilesystemSyscallKind.WRITE, 'Bad FilesystemSyscallKind'
+            return self.write_fs_jail
+
+    def main_access_check(
+        self, syscall: str, fs_jail_getter: FSJailGetter, dirfd_getter: DirFDGetter, file_reg: int
```
Xyene, line 338:
```suggestion
        self, syscall: str, *, fs_jail_getter: FSJailGetter, dirfd_getter: DirFDGetter, file_reg: int
```

dmoj/cptbox/isolate.py
```diff
@@ -201,176 +202,290 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+                e.args[0],
+            )
+        except UnicodeDecodeError as e:
+            raise AccessFailure(
+                ACCESS_ENOENT,
+                logging.WARNING,
+                'Invalid unicode: %r',
+                e.object,
+            )
+        else:
```
Xyene, line 434: Can get rid of the `else`

dmoj/cptbox/isolate.py
Riolku: Because it is a DirFDGetter, and those return ints.

dmoj/cptbox/isolate.py
Riolku: Supposed to be a decorator, but removed entirely after this new push.

Riolku:
Converting to draft because I'm not sure this will fix all our problems and I'm pending on a good solution.

dmoj/cptbox/isolate.py
```diff
@@ -201,110 +208,169 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
                     sys_minherit: ALLOW,
                     sys_thr_set_name: ALLOW,
                     sys_sigfastblock: ALLOW,
-                    sys_realpathat: self.check_file_access_at('realpathat', FilesystemSyscallKind.READ),
+                    sys_realpathat: self.handle_file_access_at(
+                        'faccessat', FilesystemSyscallKind.READ, dir_reg=0, file_reg=1
```
Xyene, line 212: This isn't `faccessat`.

dmoj/cptbox/isolate.py
```diff
@@ -56,22 +57,28 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+                sys_faccessat: self.handle_file_access_at(
+                    'faccessat', FilesystemSyscallKind.READ, dir_reg=0, file_reg=1
+                ),
+                sys_faccessat2: self.handle_file_access_at(
+                    'faccessat', FilesystemSyscallKind.READ, dir_reg=0, file_reg=1
+                ),
+                sys_access: self.handle_file_access('access', FilesystemSyscallKind.READ, file_reg=0),
+                sys_readlink: self.handle_file_access('readlink', FilesystemSyscallKind.READ, file_reg=0),
+                sys_readlinkat: self.handle_file_access_at(
+                    'faccessat', FilesystemSyscallKind.READ, dir_reg=0, file_reg=1
```
Xyene, line 71: Also not `faccessat`.

dmoj/cptbox/isolate.py
```diff
@@ -56,22 +57,28 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
-                sys_tgkill: self.do_kill,
-                sys_kill: self.do_kill,
-                sys_prctl: self.do_prctl,
+                sys_openat: self.handle_openat('openat', dir_reg=0, file_reg=1, flag_reg=2),
+                sys_open: self.handle_open('open', file_reg=0, flag_reg=1),
+                sys_faccessat: self.handle_file_access_at(
+                    'faccessat', FilesystemSyscallKind.READ, dir_reg=0, file_reg=1
+                ),
+                sys_faccessat2: self.handle_file_access_at(
+                    'faccessat', FilesystemSyscallKind.READ, dir_reg=0, file_reg=1
```
Xyene, line 66: Kind of `faccessat`? But it's `faccessat2`.

Riolku: Lol, fixed.

dmoj/cptbox/isolate.py
Riolku: Fixed.

dmoj/cptbox/compiler_isolate.py
```diff
@@ -0,0 +1,163 @@
...
+
+                debugger.on_return(on_return)
+                return True
+
+        # Emulate https://github.com/torvalds/linux/blob/v5.14/fs/utimes.c#L142-L143
+        if debugger.uarg0 != AT_FDCWD and not debugger.uarg1:
+            path = self._getfd_pid(debugger.tid, debugger.uarg0)
+            return True if self.write_fs_jail.check(path) else ACCESS_EPERM(debugger)
+
+        return self.handle_file_access_at('utimensat', FilesystemSyscallKind.WRITE, dir_reg=0, file_reg=1)(debugger)
```
Xyene, line 159: This might raise, but there's no handling for that around this function.

Xyene (commented):
Probably the exception handler wrapping should happen in `__setitem__`.

Riolku:
My main issue with `__setitem__` is that it's not clear what we're getting: are we getting a checker that will return None or raise? Are we getting a callback that returns a boolean? `ALLOW` is an integer.

Also, how sure are we that `update()` calls `__setitem__`? Should we use `__getitem__`? I feel like that's kind of painful overall. Is there somewhere sane to move this code to?

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
+
+        # normpath doesn't strip leading slashes
         projected = normalized = '/' + os.path.normpath(file).lstrip('/')
+        proc_dir = f'/proc/{debugger.tid}'
         if normalized.startswith('/proc/self'):
-            file = os.path.join(f'/proc/{debugger.tid}', os.path.relpath(file, '/proc/self'))
-            projected = '/' + os.path.normpath(file).lstrip('/')
-        elif normalized.startswith(f'/proc/{debugger.tid}/'):
+            projected = os.path.join(proc_dir, os.path.relpath(projected, '/proc/self'))
+        elif normalized.startswith(proc_dir):
```
Xyene, line 357: This changes behavior to no longer require the trailing slash

dmoj/cptbox/compiler_isolate.py
```diff
@@ -0,0 +1,150 @@
...
+
+                debugger.on_return(on_return)
+
+        else:
+            # Emulate https://github.com/torvalds/linux/blob/v5.14/fs/utimes.c#L142-L143
+            if debugger.uarg0 != AT_FDCWD and not debugger.uarg1:
+                path = self._getfd_pid(debugger.tid, debugger.uarg0)
+                if not self.write_fs_jail.check(path):
+                    raise AccessFailure(ACCESS_EPERM, 'Denying access to %s', path)
+
```
Xyene, line 143:
```suggestion

```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
-            return file, ACCESS_ENOENT
-        else:
-            if not same:
-                log.warning(
-                    'Denying access due to suspected symlink trickery: normalizes to: %s, actually: %s',
-                    normalized,
-                    real,
-                )
-                return file, ACCESS_EACCES
+            raise AccessFailure(ACCESS_ENOENT, 'Cannot stat, file: %s, projected: %s, real: %s', file, projected, real)
```
Xyene, line 366:
```suggestion
            raise AccessFailure(ACCESS_ENOENT, f'Cannot stat, {file=}, {projected=}, {real=}')
```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
+
+        if not same:
+            raise AccessFailure(
+                ACCESS_EACCES, 'Suspected symlink trickery, file: %s, projected: %s, real: %s', file, projected, real
+            )
 
         if not fs_jail.check(normalized):
-            return normalized, ACCESS_EACCES
+            raise AccessFailure(ACCESS_EACCES, 'Denying %s, normalized to %s', file, normalized)
```
Xyene, line 374:
```suggestion
            raise AccessFailure(ACCESS_EACCES, 'Denying {file}, {normalized=}')
```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
-                    'Denying access due to suspected symlink trickery: normalizes to: %s, actually: %s',
-                    normalized,
-                    real,
-                )
-                return file, ACCESS_EACCES
+            raise AccessFailure(ACCESS_ENOENT, 'Cannot stat, file: %s, projected: %s, real: %s', file, projected, real)
+
+        if not same:
+            raise AccessFailure(
+                ACCESS_EACCES, 'Suspected symlink trickery, file: %s, projected: %s, real: %s', file, projected, real
```
Xyene, line 370:
```suggestion
                ACCESS_EACCES, 'Suspected symlink trickery, {file=}, {projected=}, {real=}'
```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
-        return normalized, None
-
-    def get_full_path(self, debugger: Debugger, file: str, dirfd: int = AT_FDCWD) -> str:
-        dirfd = (dirfd & 0x7FFFFFFF) - (dirfd & 0x80000000)
-        if not file.startswith('/'):
-            dir = self._getcwd_pid(debugger.tid) if dirfd == AT_FDCWD else self._getfd_pid(debugger.tid, dirfd)
-            file = os.path.join(dir, file)
-        file = '/' + os.path.normpath(file).lstrip('/')
-        return file
+                raise AccessFailure(ACCESS_EACCES, 'Denying %s, real path %s', file, real)
```
Xyene, line 381:
```suggestion
                raise AccessFailure(ACCESS_EACCES, 'Denying {file}, real path {real}')
```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
 
-    def do_kill(self, debugger: Debugger) -> bool:
+    def handle_kill(self, debugger: Debugger) -> None:
         # Allow tgkill to execute as long as the target thread group is the debugged process
         # libstdc++ seems to use this to signal itself, see <https://github.com/DMOJ/judge/issues/183>
-        return True if debugger.uarg0 == debugger.pid else ACCESS_EPERM(debugger)
+        if debugger.uarg0 != debugger.pid:
+            raise AccessFailure(ACCESS_EPERM, 'Cannot kill other processes.')
```
Xyene, line 387:
```suggestion
            raise AccessFailure(ACCESS_EPERM, 'Cannot kill other processes')
```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
-    def do_prctl(self, debugger: Debugger) -> bool:
+    def handle_prctl(self, debugger: Debugger) -> None:
         PR_GET_DUMPABLE = 3
         PR_SET_NAME = 15
         PR_GET_NAME = 16
         PR_SET_THP_DISABLE = 41
         PR_SET_VMA = 0x53564D41  # Used on Android
-        return debugger.arg0 in (PR_GET_DUMPABLE, PR_SET_NAME, PR_GET_NAME, PR_SET_THP_DISABLE, PR_SET_VMA)
+        if debugger.arg0 not in (PR_GET_DUMPABLE, PR_SET_NAME, PR_GET_NAME, PR_SET_THP_DISABLE, PR_SET_VMA):
+            raise AccessFailure(protection_fault, 'Invalid prctl argument')
```
Xyene, line 400:
```suggestion
            raise AccessFailure(protection_fault, 'Not-whitelisted prctl option')
```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
+            return failure.handle(debugger)
+
+    return inner
+
+
+def protection_fault(self, debugger: Debugger) -> bool:
+    return False
+
+
+class AccessFailure(Exception):
```
Xyene, line 430: Nit: this is no longer just an "access" failure, as it's used for disallowing other syscalls. Probably `DeniedSyscall` or something would be better? Worth updating the log message too.

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
+    def __init__(self, handler: Callable, reason_format: str, *reason_args):
+        self.handler = handler
+        self.reason_format = reason_format
+        self.reason_args = reason_args
+
+    def handle(self, debugger: Debugger) -> bool:
```
Xyene, line 436: Do we need this function? Instead of `exn.handle(...)`, a caller can already do `exn.handler(...)`.

dmoj/executors/base_executor.py
```diff
@@ -54,6 +54,8 @@
         ExactFile('/dev/hv_tsc'),
         RecursiveDir('/dev/fd'),
     ]
+
+    BASE_WRITE_FILESYSTEM += [ExactFile('/dev/stderr')]
```
Xyene, line 58: Worth adding a comment: https://github.com/DMOJ/judge-server/commit/08791138854cdb5c1f131e7f3d496cdff1ccb80b

dmoj/cptbox/isolate.py
```diff
@@ -201,110 +204,141 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
 
-        try:
-            file = self.get_full_path(debugger, rel_file, dirfd)
-        except UnicodeDecodeError:
-            log.exception('Unicode decoding error while opening relative to %d: %r', dirfd, rel_file)
-            return '(undecodable)', ACCESS_EINVAL
+        return file
+
+    # intentionally non-normalized
+    def get_full_path(self, debugger: Debugger, rel_file: str, *, dirfd: int) -> str:
```
Xyene, line 330:
```suggestion
    def get_full_path_unnormalized(self, debugger: Debugger, rel_file: str, *, dirfd: int) -> str:
```

dmoj/cptbox/isolate.py
```diff
@@ -201,110 +204,141 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
-    def _fs_jail(self, read_open_flags, kind):
-        if kind == FilesystemSyscallKind.WRITE:
-            return lambda _: self.write_fs_jail
-        elif kind in (FilesystemSyscallKind.READ, FilesystemSyscallKind.FSTAT):
-            return lambda _: self.read_fs_jail
-        elif kind == FilesystemSyscallKind.OPEN:
-            return lambda debugger: self._fs_jail_from_open_flags(read_open_flags(debugger))
-        else:
-            raise ValueError('bad FilesystemSyscallKind value')
+    def handle_file_access(self, kind: FilesystemSyscallKind, *, file_reg: int) -> AccessChecker:
```
Xyene, line 214: Nit: move all the handlers below the definitions of the `_fs_jail_getter_from_kind` etc. helpers so that reading this code is more linear.

dmoj/cptbox/isolate.py
Xyene: It can generate some non-normalized paths now:

```python

In [5]: proc_dir = '/proc/123'

In [6]: normalized = '/proc/123'

In [7]: os.path.join('/proc/self', os.path.relpath(normalized, proc_di
   ...: r))
Out[7]: '/proc/self/.'
```

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
+    def __init__(self, handler: Callable, reason_format: str, *reason_args):
+        self.handler = handler
+        self.reason_format = reason_format
+        self.reason_args = reason_args
+
+    def handle(self, debugger: Debugger) -> bool:
+        return self.handler(debugger)
+
+    def log(self, syscall: int):
+        log.warning(
```
Xyene, line 440: Let's bump this down to `debug`.

dmoj/cptbox/isolate.py
```diff
@@ -201,110 +204,141 @@ def __init__(self, read_fs, write_fs=None, writable=(1, 2)):
...
+                if open_flags & flag == flag:
+                    return self.write_fs_jail
+
+            return self.read_fs_jail
+
+        return getter
+
+    def _fs_jail_getter_from_kind(self, kind) -> FSJailGetter:
+        def getter(debugger: Debugger) -> FilesystemPolicy:
+            if kind == FilesystemSyscallKind.READ:
```
Xyene, line 284:
```python
return {
  FilesystemSyscallKind.READ: self.read_fs_jail,
  FilesystemSyscallKind.WRITE: self.write_fs_jail
}[kind]
```

then you don't need to worry about the assert.

dmoj/cptbox/isolate.py
```diff
@@ -314,63 +348,97 @@ def _file_access_check(
...
             # If the child process uses /proc/getpid()/foo, set the normalized path to be /proc/self/foo.
             # Access rules can more easily check /proc/self.
-            normalized = os.path.join('/proc/self', os.path.relpath(file, f'/proc/{debugger.tid}'))
-        real = os.path.realpath(file)
+            normalized = os.path.join('/proc/self', os.path.relpath(normalized, proc_dir))
 
+        real = os.path.realpath(projected)
         try:
-            same = normalized == real or os.path.samefile(projected, real)
+            same = projected == real or os.path.samefile(projected, real)
```
Xyene, line 364: Was this change intended? It basically means the `elif normalized.startswith(proc_dir):` branch above is dead code, because the resulting `normalized` path is never used.

dmoj/cptbox/isolate.py
Riolku: Is it better to normalize, or to include the trailing slash? Supposedly we don't want to check the literal path `/proc/self` wrong, since under the old system it won't match

dmoj/cptbox/isolate.py
Riolku: We don't yeah.

dmoj/cptbox/isolate.py
Riolku: Isn't it better to not use an f-string? Since these are all being passed to a `log` function, which might not need to output the string, it's better not to use CPU cycles, right?

dmoj/executors/base_executor.py
Riolku: `/dev/stderr` actually fixes nothing on PASCAL: I believe the reason we have sporadic TLEs now is because of the cost of outputting to the screen, which wasn't present before, because we had this logging on DEBUG.

dmoj/cptbox/isolate.py
Riolku: It is more sensible to check `projected` with `normalized`. This way, we don't waste time if a process uses a `/proc` file, like `/proc/self/maps`, which we allow. As for `normalized`, it is still used for the jail access check, so the code is not dead. This change was intended.

dmoj/cptbox/isolate.py
Xyene: I'd rather not have to think about this subtle of a change within this big of a PR, if you'd like to make a change to this logic please wait until afterwards to do so.

dmoj/cptbox/isolate.py
Xyene: The formatting is a drop in the bucket to everything else. If we cared about performance of trapped syscalls, we wouldn't be doing things in allocating Python code.

dmoj/cptbox/isolate.py
Riolku: In the spirit of not doing too much in one PR, I'll revert this.

dmoj/executors/base_executor.py
Xyene: That's hilarious.

dmoj/cptbox/isolate.py
```diff
@@ -2,22 +2,21 @@
...
     ACCESS_EFAULT,
-    ACCESS_EINVAL,
     ACCESS_ENAMETOOLONG,
     ACCESS_ENOENT,
     ACCESS_EPERM,
     ALLOW,
     ErrnoHandlerCallback,
 )
 from dmoj.cptbox.syscalls import *
+from dmoj.cptbox.syscalls import by_id
```
Xyene, line 19: This is already implied by * above.

Riolku: No it's not: `syscalls` defines `__all__`, which is what is actually imported.

Xyene (approved):
LGTM, thanks!

PR #997, opened 2022-02-27
Make accessing Debugger registers more ergonomic

Xyene:


dmoj/cptbox/_cptbox.pyx
```diff
@@ -225,6 +225,22 @@ def memory_fd_seal(int fd):
...
+cdef class ArgAccessor:
+    cdef Debugger debugger
+    cdef object base
+
+    def __cinit__(self, Debugger debugger, object base):
+        self.debugger = debugger
+        self.base = base
+
+    def __getitem__(self, reg):
+        return getattr(self.debugger, '%s%d' % (self.base, reg))
```
quantum5, line 239: Why are you doing this in `_cptbox`? Maybe you should move this into `AdvancedDebugger` instead.

quantum5 (commented):
Otherwise the idea looks fine...

Riolku:
Agree on Quantum's comment, otherwise looks good.

PR #999, opened 2022-02-28
cptbox: move logic from rule adding to constructor

Riolku:
Basically, instead of checking that rules are correctly written when they are compiled into the jail, we should just check at construction time.

### DIFF
--- a/dmoj/cptbox/filesystem_policies.py
+++ b/dmoj/cptbox/filesystem_policies.py
@@ -24,25 +24,52 @@ class File:
     pass
 
 
-class ExactFile:
+class BaseFilesystemAccessRule:
     def __init__(self, path: str):
+        path = os.path.expanduser(path)
+        assert os.path.abspath(path) == path, 'FilesystemAccessRule must specify a normalized, absolute path to rule'
         self.path = path
+        self._assert_rule_type()
+
+    def _assert_rule_type(self) -> None:
+        if self.exists():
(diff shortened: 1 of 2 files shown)


kiritofeng:
ok to test

Riolku:
This is marginally beneficial to landlock, since it checks the paths before they are even added to the jail. It doesn't end up mattering as of now because we construct the jail anyway, but it's probably better for the long term.

dmoj/cptbox/filesystem_policies.py
```diff
@@ -24,25 +24,58 @@ class File:
...
 
-class RecursiveDir:
-    access_mode = AccessMode.RECURSIVE
+    def is_realpath(self) -> bool:
+        return self.path.startswith('/proc/self') or self.realpath() == self.path  # proc/self should not be resolved
 
-    def __init__(self, path: str):
-        self.path = path
+    def __repr__(self) -> str:
+        return f'<{self.repr_name} {self.path}>'
```
quantum5, line 53: Why are you using this `repr_name` instead of just `__class__.__name__` or something?

Riolku: Wasn't aware of its existence. Pushed a new update to use it, thanks!

Riolku:
I'm confused: did you mean to merge the PR that depended on this, #1010? If so, I'll just close this (or you can).

Riolku:
I'm just going to close this. The commit has been merged anyways.

PR #1000, opened 2022-02-28
fix syscall name in log

Riolku:
.lstrip() takes a set of characters to strip, so `sys_stat` becomes `tat` under the old system. Observed in logs.

### DIFF
--- a/dmoj/cptbox/isolate.py
+++ b/dmoj/cptbox/isolate.py
@@ -427,6 +427,8 @@ class DeniedSyscall(Exception):
         self.reason = reason
 
     def log(self, syscall: int):
-        syscall_name = by_id[syscall].lstrip('sys_')
+        syscall_name = by_id[syscall]
+        if syscall_name.startswith('sys_'):
+            syscall_name = syscall_name[4:]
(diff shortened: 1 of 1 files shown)


kiritofeng:
ok to test

dmoj/cptbox/isolate.py
```diff
@@ -427,6 +427,8 @@ def __init__(self, handler: Callable, reason: str):
         self.reason = reason
 
     def log(self, syscall: int):
-        syscall_name = by_id[syscall].lstrip('sys_')
+        syscall_name = by_id[syscall]
+        assert syscall_name.startswith('sys_')
```
quantum5, line 431: Is this really worth failing over? How about just if it starts with `sys_` do `[:4]`?

Xyene: `.replace('sys_', '', 1)`?

Riolku: I think I'd rather be a bit more explicit and take quantum's suggestion.

quantum5 (approved):

PR #1001, opened 2022-02-28
declare `proc_dir` earlier and resolve exact proc_dir

Riolku:
We can declare the proc directory earlier in the code and use it. We intentionally add a trailing slash on line 352, in case for instance, the process PID is 123 and they request /proc/12345. Finally, we add a check for the literal directory /proc/getpid()


kiritofeng:
ok to test

PR #1002, opened 2022-02-28
cptbox: compare projected with real instead of normalized

Riolku:
We should compare `projected` with `real` instead of `normalized`. This is because if our file is `/proc/self/maps`, the projected will be `/proc/getpid()/maps`, and the real path is `/proc/getpid()/maps`. We do a useless `os.samefile` call in this case.

### DIFF
--- a/dmoj/cptbox/isolate.py
+++ b/dmoj/cptbox/isolate.py
@@ -354,7 +354,7 @@ class IsolateTracer(dict):
         real = os.path.realpath(file)
 
         try:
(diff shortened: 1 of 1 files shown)


kiritofeng:
ok to test

Xyene (changes requested):
This makes the `normalized` computation dead code.

Riolku:
Not true. The jail still needs to check `normalized` and not `projected`.

PR #1003, opened 2022-02-28
testsuite: add test for opening format string

Riolku:
Raised as a concern in #998. We should add an explicit test for this. Note that on seccomp, this returns EACCES, but on landlock, it will return ENOENT, and conceptually, both are reasonable.

### DIFF
--- a/testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
+++ b/testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
@@ -1,0 +1,17 @@
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <unistd.h>
+
+int main() {
+    int fd;
+    if ((fd = open("%s", O_RDONLY)) >= 0) {
+        puts("Should not open the file");
+    } else if (errno == EACCES || errno == ENOENT) { // differs between seccomp and landlock
+        puts("Hello, World!");
+        return 0;
+    } else {
+        printf("Wrong errno: %d", errno);
+    }
+    return -1;
+}
--- a/testsuite/helloworld/tests/sandbox_c_format_open/test.yml
+++ b/testsuite/helloworld/tests/sandbox_c_format_open/test.yml
@@ -1,0 +1,5 @@
+language: C
+time: 1
+memory: 65536
+source: format_open.c
+expect: AC


kiritofeng:
ok to test

testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
```diff
@@ -0,0 +1,17 @@
...
+int main() {
+    int fd;
+    if ((fd = open("%s", O_RDONLY)) >= 0) {
+        close(fd);
+        puts("Should not open the file");
+    } else if (errno == EACCES || errno == ENOENT) { // differs between seccomp and landlock
+        puts("Hello, World!");
+    } else {
+        printf("Wrong errno: %d", errno);
+    }
```
Xyene, line 16: Missing return. These may be just tests, but still...

testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
```diff
@@ -0,0 +1,17 @@
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+int main() {
+    int fd;
+    if ((fd = open("%s", O_RDONLY)) >= 0) {
+        close(fd);
```
Xyene, line 10: We're about to shut down anyway, no need to close.

testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
```diff
@@ -0,0 +1,17 @@
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
```
Xyene, line 4: Is this header used in this code?

Xyene (changes requested):

testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
Riolku: Copied from `nametoolong.c`, but fair, will change.

testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
Riolku: I'm assuming you mean string.h, and I think the answer is no.

testsuite/helloworld/tests/sandbox_c_format_open/format_open.c
Riolku: I've added return codes based on what happened.

Xyene (approved):
LGTM, thanks!

PR #1004, opened 2022-02-28
executors: remove all regex path checks

Riolku:
Remove legacy code that added regex strings to the allowed filesystem. Disclaimer: we can't support the regex that OBJC had before, but we can make a best attempt.


kiritofeng:
ok to test

kiritofeng:
ok to test

PR #1006, opened 2022-02-28
cptbox: remove legacy PTBOX_HANDLER_STDOUTERR

Riolku:
This is from a time when we trapped `write`, and checked the arguments of it, neither of which is true now.

### DIFF
--- a/dmoj/cptbox/handlers.py
+++ b/dmoj/cptbox/handlers.py
@@ -5,7 +5,6 @@ from dmoj.cptbox._cptbox import Debugger
 DISALLOW = 0
 ALLOW = 1
 _CALLBACK = 2
-STDOUTERR = 3
 
 
 class ErrnoHandlerCallback:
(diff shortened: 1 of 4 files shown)


kiritofeng:
ok to test

quantum5 (dismissed):

quantum5 (changes requested):
Lots more references... https://github.com/DMOJ/judge-server/blob/bc88654d40a39ccc5dc24dad5127150bbc792d7b/dmoj/cptbox/handlers.py#L8 and https://github.com/DMOJ/judge-server/blob/bc88654d40a39ccc5dc24dad5127150bbc792d7b/dmoj/cptbox/handlers.py#L8 needs to be removed.

Riolku:
You say lots more, but are these the only two?

quantum5 (approved):
LGTM assuming tests pass

PR #1008, opened 2022-03-01
cptbox: add typing and fix typing on rule list

Riolku:
There were some missing types in the isolate tracer. Since we need only iterate through the list in the Policy constructor, it need not be a list.


Xyene (approved):
LGTM, thanks!

PR #1009, opened 2022-03-01
executors: fix typing on `compiler_fs`

Riolku:
We rely on the `fs` being a list in order to add it to other lists, so we should type it accordingly. MyPy complains about typing in some executors unless they are explicitly typed, see https://mypy.readthedocs.io/en/stable/common_issues.html#variance.

### DIFF
--- a/dmoj/cptbox/compiler_isolate.py
+++ b/dmoj/cptbox/compiler_isolate.py
@@ -1,8 +1,9 @@
 import struct
 import sys
+from typing import List
 
 from dmoj.cptbox._cptbox import AT_FDCWD, Debugger
-from dmoj.cptbox.filesystem_policies import ExactFile, FilesystemPolicy, RecursiveDir
+from dmoj.cptbox.filesystem_policies import ExactFile, FilesystemAccessRule, FilesystemPolicy, RecursiveDir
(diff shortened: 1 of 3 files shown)


Riolku:
We don't actually rely on these being lists right now... since we don't add them to anything (which is why it typechecks of course), but it's worth adding this for potential future use.

kiritofeng:
Should we also change https://github.com/DMOJ/judge-server/blob/master/dmoj/cptbox/isolate.py#L45 to `List`?

dmoj/cptbox/compiler_isolate.py
```diff
@@ -14,7 +15,7 @@
 
 
 class CompilerIsolateTracer(IsolateTracer):
-    def __init__(self, *, tmpdir, read_fs, write_fs):
+    def __init__(self, *, tmpdir, read_fs: List[FilesystemAccessRule], write_fs: List[FilesystemAccessRule]):
```
kiritofeng, line 18: I know this is technically out of scope of the PR, but can you also add the type annotation for `tmpdir`? (IIRC it should be `str`).

Riolku:
> Should we also change https://github.com/DMOJ/judge-server/blob/master/dmoj/cptbox/isolate.py#L45 to `List`?

Updated.

kiritofeng (approved):
LGTM, thanks!

PR #1010, opened 2022-03-05
cptbox: don't allow paths that don't exist

Riolku:
This is mostly for compatibility with landlock, which enforces this. Builds on top of #999.

### DIFF
--- a/dmoj/executors/compiled_executor.py
+++ b/dmoj/executors/compiled_executor.py
@@ -81,6 +81,9 @@ class CompiledExecutor(BaseExecutor, metaclass=_CompiledExecutorMeta):
     compiler_read_fs: Sequence[FilesystemAccessRule] = []
     compiler_write_fs: Sequence[FilesystemAccessRule] = []
 
+    # List of directories required by the compiler to be present, typically for writing to
+    compiler_required_dirs: List[str] = []
+
     def __init__(self, problem_id: str, source_code: bytes, *args, **kwargs) -> None:
         super().__init__(problem_id, source_code, **kwargs)
         self.warning = None
@@ -94,6 +97,13 @@ class CompiledExecutor(BaseExecutor, metaclass=_CompiledExecutorMeta):
         self._code = self._file(self.source_filename_format.format(problem_id=problem_id, ext=self.ext))
         with open(self._code, 'wb') as fo:
             fo.write(utf8bytes(source_code))
+
+        for path in self.compiler_required_dirs:
+            path = os.path.expanduser(path)
+            try:
+                os.mkdir(path, mode=0o775)
+            except FileExistsError:
+                pass
 
     def get_compile_args(self) -> List[str]:
         raise NotImplementedError()
--- a/dmoj/cptbox/filesystem_policies.py
+++ b/dmoj/cptbox/filesystem_policies.py
@@ -24,25 +24,52 @@ class File:
     pass
 
 
-class ExactFile:
+class BaseFilesystemAccessRule:
     def __init__(self, path: str):
+        path = os.path.expanduser(path)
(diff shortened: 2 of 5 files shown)


dmoj/executors/compiled_executor.py
```diff
@@ -95,6 +98,14 @@ def create_files(self, problem_id: str, source_code: bytes, *args, **kwargs) ->
         with open(self._code, 'wb') as fo:
             fo.write(utf8bytes(source_code))
 
+        for path in self.compiler_required_dirs:
+            path = os.path.expanduser(path)
+            assert path == os.path.abspath(path), 'compiler_required_dirs must contain only absolute normalized paths'
```
Xyene, line 103: Wait but... they don't? `~/.cache` isn't an absolute path. I don't think it's worth having any restriction here.

Xyene:
Are you sure Rust doesn't need a cache directory? What does it do when it doesn't have one?

dmoj/executors/compiled_executor.py
Riolku: What happens if the path is not absolute after `expanduser`? What do we do?

Xyene: It's fine, these are not adversarial conditions. We specify the paths.

Riolku:
Do you mean `~/.cargo`? I'm not sure. I wonder if that's created on Rust installation or something. I'll add it to the list anyway.

dmoj/executors/compiled_executor.py
Riolku: Should I change the error message? I'm very hesitant to allow non-absolute paths.

Xyene:
I don't think so, but it's worth checking? I know the first compile during autoconf takes a long time, because it has to download a bunch of crates that then get cached.

dmoj/executors/compiled_executor.py
Xyene: Why? What are you protecting against? This code reads to me like having a check for the sake of having a check, it doesn't provide utility.

Riolku:
RUST passes in CI under this patch. I'm not really sure what it's doing though, and I was having trouble testing this locally. I think docker is re-using my old volumes, or something. Do you have any tips for how I should go about testing this?

dmoj/executors/compiled_executor.py
Riolku: I've changed the error message.

Riolku: We have sanity checks in our FilesystemAccessRule, right? Why not here?

Xyene: FilesystemAccessRule can be specified by users who may not know what they're doing in `extra_fs`, these can't

Riolku: I've removed the check, but I'm still unsure where the line is for making these checks. Don't we specify the paths for the rules too? I suppose not _always_.

Xyene:
Try running `docker prune -a` then rerunning `make judge-tier${whatevertierrustisin}`.

Riolku:
Rust actually has `.cargo` setup in Docker container initialization, so we're ok.

Riolku:
Actually, if we aren't in Docker, we should create .cargo right? Is it worth adding then?

Xyene:
I don't want to particularly support running outside of Docker, it's too much of a hassle. Users are on their own if they do so anyway.

Xyene:
Specifically: Cargo installs to `.cargo`, so if they don't have `.cargo` they probably don't have Cargo installed (or have it at some weird path). Not our problem in either case.

Xyene (approved):
LGTM

PR #1011, opened 2022-03-05
cptbox: strip carriage returns in TEXT

### LINKED ISSUE #965: Text Executor Carriage Returns
It seems that since the source code is taken directly for the text executor, the output can contain carriage returns. This is probably undesirable behaviour.

Riolku:
Closes #965. Refactored the code into a mixin for future use. ZIG also currently does this.


Xyene (approved):
LGTM, thanks

Xyene:
Actually, but too late now: this wasn't a cptbox change, so it should've been tagged `executors:`

Riolku:
Ah, sorry about that. Kinda used to `cptbox` tagging now.

PR #1012, opened 2022-03-05
Report context switches and runtime version to bridge

quantum5:
The first commit simplifies the interface of `get_runtime_versions` so that no mypy asserts are needed.


Riolku (commented):
This looks fine to me.

Xyene (approved):
LGTM

PR #1013, opened 2022-03-05
judge: add support for test case prerequisites

Riolku:
Prerequisites are indicated in the YML, and they indicate that in order for a batch to run, a given set of other batches must have completed successfully.


Xyene (commented):
Let's not call these prerequisites, because that is a painful word to type.

Suggestion: `depends_on: [...]` or `requires: [...]`.

Riolku:
I like `requires`. Let me change that.

Riolku:
I'm going to use `requires` and `requirements`

Riolku:
Actually, I'm going to use `requirements` because `batch.requires` looks awkward.

dmoj/judge.py
```diff
@@ -458,21 +458,28 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
 
         case_number = 0
         is_short_circuiting = False
         is_short_circuiting_enabled = self.submission.short_circuit
+        passed_batches: Set[int] = set()
         for batch_number, cases in groupby(flattened_cases, key=itemgetter(0)):
             if batch_number:
                 yield IPC.BATCH_BEGIN, (batch_number,)
 
+                reqs = batch_requirements[batch_number - 1]  # List is zero-indexed
```
Xyene, line 479: Nit: don't abbreviate words in variable names.

```suggestion
                requirements = batch_requirements[batch_number - 1]  # List is zero-indexed
```

Riolku: Fixed, also did this elsewhere (in the validation loop).

dmoj/problem.py
```diff
@@ -226,6 +226,14 @@ def __init__(self, batch_no, config, problem, cases):
...
             raise InvalidInitException('nested batches')
         self.problem = problem
+        if any(requirement >= batch_no for requirement in self.requirements):
+            raise InvalidInitException('requirements depends on non-earlier batch')
+        if any(requirement < 1 for requirement in self.requirements):
+            raise InvalidInitException('requirements must be positive integers')
+
+    @property
+    def requirements(self):
+        return self.config.get('requirements', [])
```
Xyene, line 236: You can add `'requirements;: []` to the default `ProblemConfig` and not need this property.

Riolku: Seems to not work, since this is a batch not a ProblemConfig.

Xyene: You can do `self.requirements = config.requirements`, as is done for points.

Riolku: I'm not thoroughly convinced this is sensible, because this would do weird things if someone specified at the top level, but I've pushed this anyway.

Xyene: The behavior is identical to your original code, because of how `get` is implemented on `ConfigNode`.

Xyene (approved):
LGTM

Riolku:
If someone happens to forget to indent the requirements in a batch, they'll get a cryptic error about requirements depending on a non-earlier batch, right?

Xyene:
Only if they misindent the requirements on the very last test case. The right way to fix this is to enforce a YAML schema at load time, but it's a lot of work for little real-world benefit.

Riolku:
Why does misindenting on previous test cases not cause issues? I guess because they'd have to misindent it two levels to cause the error I'm talking about?

Xyene:
Something like

```yaml
test_cases:
- batched:
  - {in: 1.txt, out: 1.txt}
  points: 1
- batched:
  - {in: 2.txt, out: 2.txt}
  points: 1
- batched:
  - {in: 3.txt, out: 3.txt}
  points: 1
requirements: [1, 2]
- batched:
  - {in: 1.txt, out: 1.txt}
  points: 1
```

will simply fail to parse, because it's not valid YAML.

PR #1014, opened 2022-03-09
remove goto for failing seccomp

Riolku:
I'm assuming this just wasn't cleaned up at some point. Not a lot of point in having a `goto`, in my opinion it just makes this more confusing.

### DIFF
--- a/dmoj/cptbox/helper.cpp
+++ b/dmoj/cptbox/helper.cpp
@@ -101,7 +101,7 @@ int cptbox_child_run(const struct child_config *config) {
     scmp_filter_ctx ctx = seccomp_init(SCMP_ACT_TRACE(0));
     if (!ctx) {
         fprintf(stderr, "Failed to initialize seccomp context!");
-        goto seccomp_fail;
+        goto seccomp_init_fail;
     }
 
     int rc;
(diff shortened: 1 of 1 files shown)


Xyene:
This actually leaks the `scmp_filter_ctx`.

Proper code here should read:

```c
seccomp_load_fail:
    seccomp_release(ctx);
seccomp_init_fail:
    return PTBOX_SPAWN_FAIL_SECCOMP;
```

adjusting the `goto` labels as appropriate.

Riolku:
Are you strongly in favour of `goto`? I generally don't really like it. Could we refactor this to _not_ use `goto`?

Xyene:
It's idiomatic C for handling resource cleanup in a fallthrough manner, to avoid duplicating cleanup code.

Xyene (approved):
LGTM

Riolku:
BSD CI seems... broken. The error seems to be on setup of the box.

Xyene:
Try force-pushing an empty commit to get it to rerun. The box ran out of inodes. There's a leak causing directories to not get cleaned up, and there were some tens of thousands of directories in `/tmp`.

Xyene:
Eh whatever, I'll just merge.

PR #1015, opened 2022-03-09
Re-comment logic for new fallback scheme

Riolku:
This changed recently IIRC. We should fix these outdated comments.


Xyene (approved):
LGTM

PR #1017, opened 2022-03-09
executors: add compiler syscalls option to CompiledExecutor

Riolku:


dmoj/executors/compiled_executor.py
```diff
@@ -104,6 +106,20 @@ def get_compile_env(self) -> Optional[Dict[str, str]]:
     def get_compile_popen_kwargs(self) -> Dict[str, Any]:
         return {}
 
+    def _add_compiler_syscalls(self, sec: CompilerIsolateTracer) -> CompilerIsolateTracer:
```
Xyene, line 109: This is a copy of `_add_syscalls`. Let's just make that take the list of syscalls to add, and use it here.

Xyene (changes requested):

dmoj/executors/compiled_executor.py
Riolku: I completely misunderstood, let me try again.

Riolku: Alright, should be good now.

Xyene (approved):
LGTM

PR #1018, opened 2022-03-10
executors: increase compiler TL for Pascal

Riolku:
Pascal has been flaking in CI recently, and has a rather long compile time. Notably, re-running the build often makes it pass CI.


Xyene (approved):
LGTM

PR #1019, opened 2022-03-11
judge: rename requirements to dependencies

Riolku:
We should do this before anything builds on this. I think `dependencies` is a better name, is more succint, and well-known to programmers.


Riolku:
Ok maybe it's not more succinct since it's basically the same length.

Xyene (approved):
LGTM

PR #1020, opened 2022-03-12
judge: let problems specify custom limits for cases

Riolku:
Some notes on some potentially weird changes. Leaving a time limit of zero implies no time limit, and similarly with a memory limit of zero. In `test.py`, it's strange to fetch a ProblemConfig, since we never need the functionality that provides.


Riolku:
Out of band communication: the solution to this is to avoid dirtying the codebase, and instead use a custom grader.

PR #1021, opened 2022-03-12
judge: pass `result` to checkers

Riolku:


Xyene (changes requested):
If we're going to do this, let's just pass the whole result object

Riolku:
Hm.. but keep the rest of the kwargs for backward compatability?

Xyene:
Yes

Riolku:
Passing `result` now, updated tests and commit message.

kiritofeng (approved):
LGTM

Riolku:
Seems `black` is bricked right now. I'll check back later tonight.

Xyene (approved):
LGTM, thanks

PR #1024, opened 2022-03-16
judge: report (textually) the end of batches

Riolku:
Before this change:
image

After:
image


kiritofeng:
What do we think about using indentation instead (and otherwise keeping the current text)? cc @quantum5 and @Xyene

Riolku:
This sounds good to me

-------- Original Message --------

Xyene:
I think indentation would be slightly nicer here too, but overall think the change is good.

kiritofeng:
Superseded by #1105.

PR #1026, opened 2022-03-22
judgeenv: parse arguments outside of try

Riolku:
Although it's nonsense to spawn a judge without a model_file, it's also nonsense to parse these args in this try, because it leads to useless error messages.

Specifically, the judge name and key show up as None.


kiritofeng:
Implemented as part of #728.

PR #1028, opened 2022-03-22
executors: increase java compiler timelimit

Riolku:
Java is another language that flakes regularly in CI when there is high load.


Riolku:
Hm... do you know why #1018 and #410 were merged then?

kiritofeng:
Kotlin was also flaking in prod iirc, I am unaware of Java flaking in prod.

#1018 I was unaware of, perhaps @Xyene has different opinions from me (in which case I defer to him).

Riolku:
This was an accident, but whatever, see #1030.

PR #1029, opened 2022-03-29
Upgrade black to 22.3

Riolku:
This is required to fix compatability with Click 8.1.0. Let's upgrade instead of pinning Click.

See https://github.com/psf/black/pull/2966 and note that all lint jobs fail (see #1021).


kiritofeng (approved):
LGTM

Xyene (approved):
LGTM

PR #1030, opened 2022-03-29
testsuite: increase TL on java aplusb ac

Riolku:
Slower languages like Java tend to flake in FreeBSD CI.


Riolku:
As noted in the recent commit, the compiler is not the problem: rather, the one test case is just really slow. Increasing TL globally, while not likely ideal, avoids test duplication for platform (which is senseless), and is simplest.

kiritofeng:
retest this please

kiritofeng (approved):

PR #1031, opened 2022-04-20
Allow rseq syscall

alexwaeseperlman:
Prevents protection fault for `sys_rseq` on Linux kernel versions >=5.16.


Xyene:
ok to test

To be pedantic, it's not the kernel version that's the issue here, rather that glibc started using restartable sequences to implement `sched_getcpu`. A 5.16 kernel on Debian stable glibc runs fine, for instance.

alexwaeseperlman:
> ok to test
> 
> 
> 
> To be pedantic, it's not the kernel version that's the issue here, rather that glibc started using restartable sequences to implement `sched_getcpu`. A 5.16 kernel on Debian stable glibc runs fine, for instance.

My bad, thanks for correcting.

PR #1038, opened 2022-09-04
Type `setproctitle`

kiritofeng:
Fix `All conditional function variants must have identical signatures`, see https://github.com/python/mypy/issues/10575.


Riolku (approved):
LGTM

Xyene (approved):
LGTM, thanks!

PR #1040, opened 2022-10-02
Clean up cache when worker finishes

### LINKED ISSUE #684: Compiled executor caching is broken
Regressed in https://github.com/DMOJ/judge-server/pull/673, since now the cache is per-worker-process, and gets destroyed on exit. Will require a rewrite of the caching system.

kiritofeng:


Xyene (approved):
LGTM, thanks!

PR #1051, opened 2022-10-02
Print versions during autoconfig; fixes #644

### LINKED ISSUE #644: dmoj-autoconf should print runtime versions
...the same way we do during self-testing after #545.

kiritofeng:


kiritofeng:
image
Picture; note that entries are not left aligned because file paths can be very long.

kiritofeng:
New picture
image

Xyene (approved):
LGTM if CI passes, thanks.

PR #1054, opened 2022-10-02
Print seccomp errors to parent stderr

quantum5:
This avoids non-fatal seccomp errors polluting stderr in generators.


Xyene (approved):
Good catch, thanks

PR #1055, opened 2022-10-02
Make `INT_MAX_NUMBER_DIGITS` backwards compatible

kiritofeng:
> when you remember we don't run the latest Python in prod


Xyene (approved):
LGTM, thanks!

PR #1056, opened 2022-10-02
Use /code paths for python in arm64 CI

quantum5:
This is because our python builds expect the headers in `/code`.

Instead of mounting the tree in /code and copying it to `/judge` where it is executed, we now mount it in `/source` and copy it to `/code` and execute it there.


kiritofeng (approved):
Seems legit

PR #1057, opened 2022-10-03
executors: merge Java executors

Xyene:
We had multiple executors that looked the same, for different Java versions. This caused unnecessary churn -- we just need to keep Java 8 around and separate because it remains the fastest, but all other Java executors can share the same name.


dmoj/executors/JAVA.py
```diff
@@ -2,9 +2,9 @@
 
 
 class Executor(JavacExecutor):
-    compiler = 'javac9'
-    vm = 'java9'
-    jvm_regex = r'java-9-|openjdk9|oracle-java9'
+    compiler = 'javac'
+    vm = 'java'
+    jvm_regex = r'(?:java-|openjdk)(?:9|1[0-9])'
```
quantum5, line 7:
```suggestion
    jvm_regex = r'(?:java-|openjdk)(?:9|[1-9][0-9])'
```

Let's just do this so we don't have to do any work when 20 comes around.

PR #1062, opened 2022-10-14
executors/RUST: update package versions

Riolku:
This will definitely fail in CI. The reason is that [this file](https://github.com/DMOJ/runtimes-docker/blob/master/tier2/Dockerfile#L26-L29) is responsible for the fetching of RUST packages. Since it always fetches from master, only when this PR is merged can CI pass.

To make sure it will pass, I tested locally with:

```
cp ~/Projects/dmoj/judge/judge-server/dmoj/executors/RUST.py . && sed '/^CARGO_TOML/,/^"""/!d;//d' RUST.py > Cargo.toml && sed '/^CARGO_LOCK/,/^"""/!d;//d' RUST.py > Cargo.lock && mkdir -p src && sed '/^HELLO_WORLD_PROGRAM/,/^"""/!d;//d' RUST.py > src/main.rs && cargo fetch
```

Essentially, I do the same thing as the Dockerfile does, fetch the new packages, and then run cargo offline, like it is in Docker:

```
DMOJ_CARGO_OFFLINE=y python3 .docker.test.py
```

This worked for me, so I believe this PR should pass when merged.

I didn't upgrade `lazy_static` because it is a dependency of the [dmoj-rust](https://github.com/DMOJ/dmoj-rust/blob/master/Cargo.toml) package, and as such shouldn't be changed. It could be upgraded to `0.2.11`, since that has compatibility according to [the manifest spec](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio), but that seems pointless. `libc` and `rand` are the actual important packages. `libc`'s upgrade should cause no incompatibility, and `rand` is so outdated that it's practically unusable these days.

## Breaking Changes

Any rust subs using `rand` are likely to break if rejudged, but there's not a whole lot to be done here...

### DIFF
--- a/dmoj/executors/RUST.py
+++ b/dmoj/executors/RUST.py
@@ -11,7 +11,9 @@ version = "1.0.0"
 
 [dependencies]
 dmoj = "0.1"
-rand = "0.3"
+libc = "0.2"
+rand = "0.8"
+rand_xoshiro = "0.6"
 
 [profile.release]
 strip = "symbols"


Riolku:
#1065 should be merged first (this PR should be rebased on it).

Riolku:
New changes after rebase:

`libc` and `rand_xoshiro` are part of this PR now. We need `libc` because in particular for interactors, we want to close `stdout` to signal that we are done communicating, and so we can check that we get EOF from the interactor (or process; whichever side we're one).

`rand_xoshiro` is the [recommended PRNG](https://docs.rs/rand/latest/rand/rngs/struct.SmallRng.html) by rand. We want it because we need a _reproducible_ PRNG. It is easy to imagine a case where `rand::SmallRng` changes its algorithm, and with it the data for a handful of problems. Adding this package remedies the problem. It is more mature than `rand_chacha` as well.

## Slippery Slope

Rust's standard library is smaller than others, and because of that the community is used to using crates to get the functionality they need.

As an alternative, we could expose, in DMOJ/dmoj-rust, the functionality we want. This would allow us to choose where and what is exported. However, users would not be able to use the default documentation, and we would have to document what we added.

## Testing

This fails in CI because `rand` uses some new dependencies since `0.3`, but the docker image pulls the TOML from master, not this PR (and changing this behaviour seems non-trivial), so the required dependencies don't exist (and can't be pulled since we use `--offline`, which I don't think we should change).

However, it is not overly challenging to rebuild the docker images, refetching the correct dependencies, and running the tests from there. This is what I did locally, and it worked: so I am inclined to believe that we are good to go.

Xyene (approved):
LGTM, thanks!

PR #1063, opened 2022-10-14
Rust allow libc

Riolku:
Based on #1062.

Allow `libc` in user submissions. One clear use case of this is that in interactive problems, it is desirable to close `stdout` to signal to the interactor that interaction is over (or as the interactor, to signal the same thing). Rust obviously won't let us do this without unsafe code, but without `libc`, it's not even possible.

This signalling is useful because, for instance, validators want to ensure that the interactor doesn't output anything strange after `stdout` is closed, but it's more useful for interactors to make sure that the user doesn't output anything else after the answer is received (which, under a strict definition, should result in a WA verdict).

### DIFF
--- a/dmoj/executors/RUST.py
+++ b/dmoj/executors/RUST.py
@@ -11,7 +11,7 @@ version = "1.0.0"
 
 [dependencies]
 dmoj = "0.1"
-rand = "0.3"
+rand = "0.8"
 """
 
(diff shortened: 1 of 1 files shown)


Riolku:
#1065 should merge first.

Riolku:
Superseded by #1062. We have to do all these at once, so let's stuff it into one PR.

Issue #1060, opened 2022-10-04
Support .NET (Core)


timbussmann:
With .NET Core being supported on Linux, it would be great to use the actual .NET runtime instead of Mono. The latest LTS .NET version (6.0) would bring massive performance benefits and many new useful language features. Writing code for Mono is a lot more effort compared to just using .NET nowadays, both on Windows and Linux.

One major issue with .NET is that it's very difficult to compile single code files. Compiling a C# file usually requires a project file, otherwise, the `dotnet` commands will fail, and bypassing this with direct use of the compiler is brittle. The project file could easily be hard-coded or generated on the fly though.

I'd be happy to assist with this but given my lack of experience with the server implementation and Python, I'm struggling to get started. I've already proposed [this PR](https://github.com/DMOJ/runtimes-docker/pull/40) to add the .NET 6 SDK to the runtimes image.

Riolku:
Rust uses some amount of a project file (it has a Cargo.toml hardcoded), so maybe we could do something like that for .NET as well.

timbussmann:
Thanks for the hint with the Rust executor @Riolku , that helped to get me started. I'm struggling to grasp how `get_versionable_commands` is working, is there any documentation that you can point me to?

Riolku:
Go read `base_executor.py`. Seems the command returns the canonical name and path of all parts relevant to a runtime's version.

timbussmann:
I've started a first draft via https://github.com/DMOJ/judge-server/pull/1064, help and feedback are very welcome.

PR #1064, opened 2022-10-14
.NET 6 support

timbussmann:
relates to https://github.com/DMOJ/judge-server/issues/1060
requires .NET 6 available on the runtimes image: https://github.com/DMOJ/runtimes-docker/pull/40

A first attempt at adding .NET 6 support. This is what I cobbled together based on the some other executor implementations.

Given my lack of detailed understanding of the judge server and python, there are most likely issues with the implementation. Still, it is somewhat complicated to test the behavior locally (especially without the updated runtime image). I'd appreciate some help/guidance on how to test the implementation locally/on docker.

### DIFF
--- a/dmoj/executors/NETCS.py
+++ b/dmoj/executors/NETCS.py
@@ -19,6 +19,7 @@ HELLO_WORLD_PROGRAM = """\
 Console.WriteLine("Hello, World!");
 """
 
+
 class Executor(CompiledExecutor):
     ext = 'cs'
     command = 'dotnet'
@@ -36,10 +37,6 @@ class Executor(CompiledExecutor):
 
         with open(self._file('DMOJ.csproj'), 'wb') as f:
             f.write(CSPROJ)
-
-    @classmethod
-    def get_versionable_commands(cls):
-        return [('dotnet', os.path.join(os.path.dirname(cls.get_command()), 'dotnet'))]
 
     def get_compile_args(self):
         return [self.get_command(), 'publish', '--configuration', 'release', '--self-contained', 'false']
--- a/dmoj/executors/NETCS.py
+++ b/dmoj/executors/NETCS.py
@@ -1,0 +1,48 @@
+import os
+
+from dmoj.cptbox.filesystem_policies import ExactFile, RecursiveDir
+from dmoj.executors.compiled_executor import CompiledExecutor
+
+CSPROJ = b"""\
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <PublishSingleFile>true</PublishSingleFile>
+  </PropertyGroup>
+</Project>
+"""
+
+HELLO_WORLD_PROGRAM = """\
+Console.WriteLine("Hello, World!");
+"""
+
+class Executor(CompiledExecutor):
+    ext = 'cs'
+    command = 'dotnet'
+    test_program = HELLO_WORLD_PROGRAM
+    compiler_time_limit = 20
+    compiler_write_fs = [
+        RecursiveDir('~/.nuget/packages'),
+        RecursiveDir('~/.local/share/NuGet'),
+        RecursiveDir('/tmp/NuGetScratch'),
+    ]
+
+    def create_files(self, problem_id, source_code, *args, **kwargs):
+        with open(self._file('Program.cs'), 'wb') as f:
+            f.write(source_code)
+
+        with open(self._file('DMOJ.csproj'), 'wb') as f:
+            f.write(CSPROJ)
+
+    @classmethod
+    def get_versionable_commands(cls):
+        return [('dotnet', os.path.join(os.path.dirname(cls.get_command()), 'dotnet'))]
+
+    def get_compile_args(self):
+        return [self.get_command(), 'publish', '--configuration', 'release', '--self-contained', 'false']
+
+    def get_compiled_file(self):
+        return self._file('bin', 'release', 'net6.0', 'DMOJ.exe')


dmoj/executors/NETCS.py
```diff
@@ -0,0 +1,45 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <PublishSingleFile>true</PublishSingleFile>
+  </PropertyGroup>
+</Project>
+"""
+
+HELLO_WORLD_PROGRAM = """\
+Console.WriteLine("Hello, World!");
+"""
```
Riolku, line 20: I know RUST does this, but since this is one line, and we don't have any of RUST's "prefetch libraries" business, let's remove this and inline it.

dmoj/executors/NETCS.py
```diff
@@ -0,0 +1,45 @@
...
+            f.write(source_code)
+
+        with open(self._file('DMOJ.csproj'), 'wb') as f:
+            f.write(CSPROJ)
+
+    def get_compile_args(self):
+        return [self.get_command(), 'publish', '--configuration', 'release', '--self-contained', 'false']
+
+    def get_compiled_file(self):
+        return self._file('bin', 'release', 'net6.0', 'DMOJ.exe')
```
Riolku, line 45: Please add a newline

Riolku, line 45: Is it still `.exe` on linux?

Riolku (changes requested):
Code doesn't look egregious. Did you setup a copy of the modified docker image and test this locally?

dmoj/executors/NETCS.py
```diff
@@ -0,0 +1,45 @@
...
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <PublishSingleFile>true</PublishSingleFile>
+  </PropertyGroup>
+</Project>
+"""
+
+HELLO_WORLD_PROGRAM = """\
+Console.WriteLine("Hello, World!");
```
Xyene, line 19: This program should echo input to output, not just write out "Hello, World!". See e.g. https://github.com/DMOJ/judge-server/blob/master/dmoj/executors/C.py#L11

Riolku: Is Rust's program wrong then? https://github.com/DMOJ/judge-server/blob/master/dmoj/executors/RUST.py#L78

Riolku: I'll update it sometime later then.

timbussmann: 👍 will change the behavior

dmoj/executors/NETCS.py
timbussmann: eh, no. Just a bad habit 😄 good catch!

timbussmann:
> Did you setup a copy of the modified docker image and test this locally?

@Riolku tbh I'm struggling a bit with this. locking at the docker files in `.docker` they seem to load files from `https://github.com/DMOJ/judge-server/archive/"${TAG}".tar.gz` (e.g. in https://github.com/DMOJ/judge-server/blob/master/.docker/tier3/Dockerfile) instead of using local files. Do I have to modify the docker file or is there an easier way to test this locally?

Xyene:
I'd say:

* Build the runtimes-docker image locally with your changes.
* Build the corresponding judge image as-is, it should base itself off your runtimes-docker build rather than pull in a copy from docker.io. As you note this'll pull in code from the master branch; this is fine for now.
* Pass `-v /path/to/your/judge/checkout/including/this/PR:/judge` when starting the Docker container. This will mount your sources over where the container expects judge sources to live, allowing you to test this PR.

timbussmann:
thanks the `-v /path/to/your/judge/checkout/including/this/PR:/judge` did the trick (although it required some extra workarounds when running on Windows)

> (Instead of editing /judge-runtime-paths.yml in the container you can also pull this trick for it as well.)

how would this work, given the `judge-runtime-paths.yml` is in the root folder?

running the container yields some errors that seem to be related to some access protection mechanism:

```
Self-testing NETCS:  Traceback (most recent call last):
  File "/judge/dmoj/cptbox/tracer.py", line 286, in _callback
    return callback(self.debugger)
  File "/judge/dmoj/cptbox/isolate.py", line 412, in inner
    check(debugger)
  File "/judge/dmoj/cptbox/isolate.py", line 298, in check
    self._access_check(debugger, full_path, fs_jail)
  File "/judge/dmoj/cptbox/isolate.py", line 375, in _access_check
    if not fs_jail.check(real):
  File "/judge/dmoj/cptbox/filesystem_policies.py", line 124, in check
    assert os.path.abspath(path) == path, 'Must pass a normalized, absolute path to check'
AssertionError: Must pass a normalized, absolute path to check
Exception ignored in: 'dmoj.cptbox._cptbox.Process._syscall_handler'
  File "/judge/dmoj/cptbox/tracer.py", line 286, in _callback
    return callback(self.debugger)
  File "/judge/dmoj/cptbox/isolate.py", line 412, in inner
    check(debugger)
  File "/judge/dmoj/cptbox/isolate.py", line 298, in check
    self._access_check(debugger, full_path, fs_jail)
  File "/judge/dmoj/cptbox/isolate.py", line 375, in _access_check
    if not fs_jail.check(real):
  File "/judge/dmoj/cptbox/filesystem_policies.py", line 124, in check
    assert os.path.abspath(path) == path, 'Must pass a normalized, absolute path to check'
AssertionError: Must pass a normalized, absolute path to check
```

any suggestions?

Riolku:
The code causing this error is... kind of wrong. I'm honestly a bit confused because a different PR will handle `/proc/self` differently. 

I guess the TL;DR is congratulations, `.NET` does funky stuff with files and it causes errors in our sandbox, so you probably can't test this right away.

Also, note that we don't support Windows anymore (how did you even test on Windows...).

timbussmann:
> The code causing this error is... kind of wrong. I'm honestly a bit confused because a different PR will handle /proc/self differently.

based on the `Must pass a normalized, absolute path to check` message I was wondering whether this indicates some incorrect file paths?

> I guess the TL;DR is congratulations, .NET does funky stuff with files and it causes errors in our sandbox, so you probably can't test this right away.

so is there a way to resolve this?

> Also, note that we don't support Windows anymore (how did you even test on Windows...). Did you use WSL or something?

well, let's say it has been a journey:
1. I've been using VS Code dev containers to have a python environment available in the first place and those containers are Linux containers that mount the source folders from Windows.
1. The docker images need to be built back on the Windows side again though since that's where Docker is installed. 
1. Now running the docker containers from windows didn't work due to mounting the the local source folders because apparently, I can't easily mount arbitrary folder paths on a Windows drive. That's where WSL came into play because WSL can access the same docker instance but the on Linux I can properly mount folders (and then refer to the the `/mnt/c/...` path to the Windows folder.
1. now I had to modify the `judge-runtime-paths.yml` and I couldn't figure out how I could easily use the suggested `-v` approach so I copied the file from container (`docker cp`), modified it and moved it back in. I think I also overwrote the entry-path on `docker run` to just start a shell and fix the file inside the container and then calling the regular `./docker/entry` script manually.
1. At this point the line endings finally came back to give me some more troubles because the mounted start script for the docker container had Windows newline characters (the source folder is still on Windows) and that made the docker container refuse the bash script until I forced windows to use Linux-style line endings.

(I guess I'm just writing this all down so I can remember how to do this again at some point 🤣 )

PR #1065, opened 2022-10-17
executors/RUST: Simplify TOML + Lock

Riolku:
`cargo build --offline` will happily generate a lock file from the packages we pulled during docker initialization time, so there's no reason to maintain this massive block of text.

### DIFF
--- a/dmoj/executors/RUST.py
+++ b/dmoj/executors/RUST.py
@@ -12,55 +12,10 @@ version = "1.0.0"
 [dependencies]
 dmoj = "0.1"
 rand = "0.3"
-"""
-
-CARGO_LOCK = b"""\
-cargo-features = ["strip"]
-
-[root]
-name = "user_submission"
-version = "1.0.0"
-dependencies = [
- "dmoj 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "rand 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
-]
 
 [profile.release]
 strip = "symbols"
-
-[[package]]
-name = "dmoj"
-version = "0.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "lazy_static 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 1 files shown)


Riolku:
You could argue that we save compilation time by generating it upfront I guess. A middle ground would be generating the lock file once (maybe in docker? maybe not?) and caching it.

But like, let's just.. not.

Riolku:
This PR can't be merged until the docker code is updated to not try and grab `Cargo.lock` from `RUST.py`.

Riolku:
Relevant docker PR is DMOJ/runtimes-docker#55.

dmoj/executors/RUST.py
```diff
@@ -12,55 +12,10 @@
...
-
-[root]
-name = "user_submission"
-version = "1.0.0"
-dependencies = [
- "dmoj 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "rand 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)",
-]
 
 [profile.release]
```
Xyene, line 16: Can you confirm that you see binaries getting stripped after this change? I see you also removed `cargo-features = ["strip"]` above.

Riolku: Sure, let me go double check.

Riolku: Here's the relevant manifest reference entry https://doc.rust-lang.org/cargo/reference/profiles.html#strip

Riolku: I get:

```
~/Projects/dmoj/judge/judge-server$ file user_submission 
user_submission: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=60849cf0f042bba9a714ff8060249b92104a97c4, for GNU/Linux 3.2.0, stripped
```

Riolku: I added the relevant `strip` section to the TOML, and it seems binaries are indeed stripped, so we should be all good.

Xyene (approved):
Alright, thanks for checking, let's see if this works.

PR #1066, opened 2022-10-18
Rust Edition & Test Program Upgrade

Riolku:
Using Rust edition 2021 allows us to avoid `extern crate`, which is old syntax (2015).

Also, the new program actually echos input.

This PR _should_ pass in CI, demonstrating that the new version works, but _should not_ be merged until a corresponding docker PR is ready. In particular, this PR and the docker PR should be merged at the same time.

### DIFF
--- a/dmoj/executors/RUST.py
+++ b/dmoj/executors/RUST.py
@@ -8,6 +8,7 @@ CARGO_TOML = b"""\
 [package]
 name = "user_submission"
 version = "1.0.0"
+edition = "2021"
 
 [dependencies]
 dmoj = "0.1"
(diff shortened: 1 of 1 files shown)


Riolku:
Edit: this should pass in CI if it were standalone, but it's not, because it depends on #1062.

dmoj/executors/RUST.py
```diff
@@ -8,74 +8,36 @@
...
-"checksum lazy_static 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)" = "6abe0ee2e758cd6bc8a2cd56726359007748fbf4128da998b65d0b70f881e19b"
-"checksum libc 0.2.18 (registry+https://github.com/rust-lang/crates.io-index)" = "a51822fc847e7a8101514d1d44e354ba2ffa7d4c194dcab48870740e327cac70"
-"checksum rand 0.3.15 (registry+https://github.com/rust-lang/crates.io-index)" = "022e0636ec2519ddae48154b028864bdce4eaf7d35226ab8e65c611be97b189d"
-"""  # noqa: E501
-
-HELLO_WORLD_PROGRAM = """\
-#[macro_use] extern crate dmoj;
-extern crate rand;
+TEST_PROGRAM = """\
+use dmoj as _;
```
Xyene, line 24: Aside from this PR, we should also update the site template for Rust code, helloworld, etc.

Are these `use`s actually necessary for runtimes-docker to pull in the right packages (as opposed to trusting our TOML), or are they just to prove that things worked correctly for CI?

Riolku: They are just to prove that stuff works for CI. We can add a comment, or remove them, or whatever.

Xyene: Let's just leave a comment. Having them seems useful.

Riolku: Should be pushed and rebased.

Xyene (approved):
LGTM, thanks!

PR #1075, opened 2022-11-05
Pretest dependencies

Riolku:
This PR changes the previous "batch dependencies" in a few ways:
1. Dependencies can now be specified on any top-level unit, case or batch.
2. Dependencies are specified with YAML alias notation, not numbers.

It also implements pretests, including a `PRETEST-END` IPC packet.


dmoj/judge.py
```diff
@@ -39,6 +37,7 @@ class IPC(Enum):
     RESULT = 'RESULT'
     BATCH_BEGIN = 'BATCH-BEGIN'
     BATCH_END = 'BATCH-END'
+    PRETEST_END = 'PRETEST-END'
```
Xyene, line 40: We should also add a `PRETEST-BEGIN` packet so that the state machine on the bridge side is trivial to write.

Riolku: Ok, only question is, if there are no pretests, are these packets sent?

testsuite/dependencies/init.yml
```diff
@@ -0,0 +1,20 @@
...
+- &p2 {in: 2.txt, out: 1.txt}
+- &p3 {in: 3.txt, out: 10.txt, dependencies: [*p1]}
+test_cases:
+- &m1 {in: 5.txt, out: 16.txt, points: 1, dependencies: [*p1, *p3]}
+- &m2
+  batched:
+  - {in: 4.txt, out: 2.txt}
+  - {in: 6.txt, out: 3.txt}
+  points: 1
+  dependencies: [*p2]
```
Xyene, line 13: Weak preference to naming this node `depends`, `dependencies` is pretty hard to type over and over by hand.

dmoj/judge.py
Xyene: Ideally they shouldn't be, but I don't think it'd change bridge-side logic much if they were. Just a small efficiency thing to not send them at all if we don't have pretests.

testsuite/dependencies/init.yml
Riolku: I don't really care either way, to be honest.

dmoj/problem.py
```diff
@@ -396,3 +392,25 @@ def __getstate__(self):
...
+
+
+class BatchedTestCase(AbstractTestCase):
+    def __init__(self, position, batch, config, problem):
+        super().__init__(position, batch, config, problem)
+
+
+class StandaloneTestCase(AbstractTestCase, TopLevelCase):
+    def __init__(self, top_level_position, position, config, problem):
+        AbstractTestCase.__init__(self, position, 0, config, problem)
```
Xyene, line 404: Do things break if we do this?

```suggestion
        AbstractTestCase.__init__(self, position, None, config, problem)
```

Riolku: I didn't test it. I'm not even sure we need this argument on `AbstractTestCase`. Admittedly, I just mirrored the old behaviour.

Xyene: Yeah, it'd be nice to either drop the batch parameter from the `AbstractTestCase` class, or ensure that `None` is valid here. Using `0` to mean "no batch" is kind of sketchy.

Xyene:
We'll need an accompanying site-side PR to not crash the bridge when we send it `PRETEST-{BEGIN,END}` packets.

https://github.com/DMOJ/online-judge/blob/f3cd24e3ecad89c2492917df9229009c2634baed/judge/bridge/judge_handler.py#L40-L54

dmoj/problem.py
Riolku: Oh I remember. The batch gets passed to checkers. https://github.com/DMOJ/judge-server/blob/master/dmoj/graders/standard.py#L61

So test cases have to have a batch field. Do you want to make it `None`? That seems better, but will it break anything?

Xyene: Let's just make it `None` and roll back if prod breaks ¯\_(ツ)_/¯

dmoj/judge.py
```diff
@@ -459,69 +467,91 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+        failed_pretests: Set[int] = set()
 
-        yield IPC.GRADING_END, ()
+        def should_skip_main_test(case: TopLevelCase, failed_so_far: Set[int]) -> bool:
+            bad_pretests = False
+            if case.dependencies is None:
+                bad_pretests = bool(failed_pretests)
+            else:
+                bad_pretests = bool(failed_pretests & set(case.dependencies))
+            return bad_pretests or should_skip_test(case, failed_so_far)
```
Riolku, line 540: Do you think these functions are easier understood if they were `should_run_test`, i.e., they returned the inverse of what they currently do?

Riolku: Updated to be `should_run_test`. I think that's a little less confusing to be honest.

Riolku:
Also have `dependencies` always be a set instead of a list that gets coerced to a set over and over.

dmoj/graders/base.py
```diff
@@ -29,37 +32,43 @@ def abort_grading(self):
             except OSError:
                 pass
 
-    def _resolve_testcases(self, cfg, batch_no=0):
-        cases = []
-        for case_config in cfg:
+    def _resolve_testcases(self, cfg) -> List[TopLevelCase]:
+        cases: List[TopLevelCase] = []
+        for top_level_position, case_config in enumerate(cfg):
```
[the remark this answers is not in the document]
Riolku: Fixed.

dmoj/judge.py
```diff
@@ -77,6 +77,12 @@ def __init__(self, packet_manager: packet.PacketManager) -> None:
         self.updater_signal = threading.Event()
         self.updater = threading.Thread(target=self._updater_thread)
 
+        self.in_pretests = False
+        self.pretest_batch = 0
```
Xyene, line 81: Where do we reset these values?

dmoj/judge.py
```diff
@@ -218,13 +235,29 @@ def _ipc_result(self, report, batch_number: Optional[int], case_number: int, res
...
+
         self.packet_manager.test_case_status_packet(case_number, result)
 
-    def _ipc_batch_begin(self, report, batch_number: int) -> None:
+    def _ipc_batch_begin(self, report, _batch_number: int) -> None:
         self.packet_manager.batch_begin_packet()
-        report(ansi_style('#ansi[Batch #%d](yellow|bold)' % batch_number))
+        if self.in_pretests:
+            self.pretest_batch += 1
+            report(ansi_style('#ansi[Pretest Batch #%d](yellow|bold)' % self.pretest_batch))
```
Xyene, line 257: Nit to match the rest of the casing

```suggestion
            report(ansi_style('#ansi[Pretest batch #%d](yellow|bold)' % self.pretest_batch))
```

dmoj/judge.py
Riolku: Moved these values to be initialized in the `grading_begin` packet.

dmoj/problem.py
```diff
@@ -396,3 +391,25 @@ def __getstate__(self):
...
+class StandaloneTestCase(AbstractTestCase, TopLevelCase):
+    def __init__(self, position, config, problem):
+        AbstractTestCase.__init__(self, position, None, config, problem)
+        TopLevelCase.__init__(self, config)
+
+
+class Batch(TopLevelCase):
+    def __init__(self, batch_no, config, problem, cases: List[BatchedTestCase]):
+        super().__init__(config)
+        self.batch = batch_no
```
Xyene, line 410: Whatever we call this variable, we should be consistent throughout.

Riolku: Done. I think `batch_no` is a much worse name. Let's just type the variable where it appears.

dmoj/judge.py
```diff
@@ -459,69 +491,92 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+                return True
+            else:
+                return len(case.dependencies & failed_so_far) == 0
+
+        failed_pretests: Set[int] = set()
 
-                yield IPC.BATCH_END, (batch_number,)
-                is_short_circuiting &= is_short_circuiting_enabled
+        def should_run_main_test(case: TopLevelCase, failed_so_far: Set[int]) -> bool:
+            good_pretests = False
```
Xyene, line 559: Don't need this variable here, it's defined in both branches.

dmoj/judge.py
```diff
@@ -459,69 +491,92 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+            if case.dependencies is None:
+                # Default: depends on all pretests.
+                good_pretests = len(failed_pretests) == 0
```
Xyene, line 562:
```suggestion
                good_pretests = not failed_pretests
```

dmoj/judge.py
```diff
@@ -459,69 +491,92 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+            else:
+                good_pretests = len(failed_pretests & case.dependencies) == 0
```
Xyene, line 564:
```suggestion
                good_pretests = not (failed_pretests & case.dependencies)
```

dmoj/judge.py
Riolku: Fixed

dmoj/judge.py
```diff
@@ -459,69 +491,91 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
                 # judge controller only cares about the trimmed output, and shouldn't waste memory buffering the full
                 # output. So, we trim it here so we don't run out of memory in the controller.
                 result.proc_output = result.output
-                yield IPC.RESULT, (batch_number, case_number, result)
+                return result
 
-            if batch_number:
-                if not is_short_circuiting:
-                    passed_batches.add(batch_number)
+        def should_run_test(case: TopLevelCase, failed_so_far: Set[int]) -> bool:
```
Xyene, line 549: Could we rename this to `should_run_pretest` and inline the important bits into `should_run_main_test`? AFAICT the first branch here is never entered in the call to `should_run_test` within `should_run_main_test`, and that'd make following the logic easier.

dmoj/judge.py
```diff
@@ -459,69 +491,125 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+                skip_current = skip_all or not should_run(test, failed_tests)
+                if isinstance(test, Batch):
+                    batch_number += 1
+                    yield IPC.BATCH_BEGIN, (batch_number,)
+
+                    if skip_current:
+                        yield from skip_batched_cases(test.batched_cases)
+                    else:
+                        batch_verdict = yield from grade_batched_cases(test.batched_cases)
+                        failed = batch_verdict == Result.WA
```
Riolku, line 556: Note to reviewer: we can't condense this AFAICT, it's actually a syntax error if we try.

dmoj/judge.py
```diff
@@ -218,13 +234,29 @@ def _ipc_result(self, report, batch_number: Optional[int], case_number: int, res
...
+        if self.in_pretests:
+            self.pretest_case += 1
+            report(
+                ansi_style(
+                    '%sPretest case %2d %-3s %s' % (case_padding, self.pretest_case, colored_codes[0], case_info)
+                )
+            )
+        else:
+            self.main_case += 1
+            report(ansi_style('%sTest case %2d %-3s %s' % (case_padding, self.main_case, colored_codes[0], case_info)))
```
Xyene, line 248: This string is almost identical to the one used for pretests. Can we make `"Test case"` and `"Pretest case"` substituted in with another `%s`?

Same comment below for batches.

dmoj/judge.py
```diff
@@ -194,11 +196,25 @@ def _ipc_compile_message(self, _report, compile_message: str) -> None:
         self.packet_manager.compile_message_packet(compile_message)
 
     def _ipc_grading_begin(self, _report, is_pretested: bool) -> None:
+        self.in_pretests = False
+        self.pretest_batch = 0
```
Xyene, line 200: Can we use a single set of batch/case variables, and reset them when we end pretests? That way we don't have dangling state after pretests finish.

dmoj/judge.py
```diff
@@ -459,69 +491,132 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
             if hasattr(binary, 'warning') and binary.warning is not None:
                 yield IPC.COMPILE_MESSAGE, (binary.warning,)
 
-        yield IPC.GRADING_BEGIN, (self.grader.run_pretests_only,)
+        class GradingAbort(Exception):
+            pass
+
+        def grade(case: AbstractTestCase) -> Result:
```
Xyene, line 497:
```suggestion
        def grade_single(case: AbstractTestCase) -> Result:
```

To match the naming of the rest of the functions.

dmoj/judge.py
```diff
@@ -459,69 +491,132 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
-                        # past).
-                        is_short_circuiting |= batch_number is not None or is_short_circuiting_enabled
+            return True
 
-                # Legacy hack: we need to allow graders to read and write `proc_output` on the `Result` object, but the
-                # judge controller only cares about the trimmed output, and shouldn't waste memory buffering the full
-                # output. So, we trim it here so we don't run out of memory in the controller.
-                result.proc_output = result.output
-                yield IPC.RESULT, (batch_number, case_number, result)
+        failed_pretests: Set[int] = set()
```
Xyene, line 588: Rather than have this state here captured by the closure in `should_run_main_test`, what if we took it as the first parameter to `should_run_main_test`, and then applied it with `partial(...`?

dmoj/judge.py
```diff
@@ -459,69 +491,132 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+        yield IPC.GRADING_BEGIN, (self.grader.run_pretests_only,)
+        pretests, main_tests = self.grader.cases()
+        try:
+            pretest_result = GradeManyResult.OK
+            if pretests:
+                yield IPC.PRETEST_BEGIN, ()
+                failed_pretests, pretest_result = yield from grade_many(pretests, should_run_pretest)
+                yield IPC.PRETEST_END, ()
+            if not self.grader.run_pretests_only:
+                if pretest_result == GradeManyResult.SKIP_ALL:
```
Xyene, line 612: Can this ever happen without `len(failed_pretests) != 0`? Do we need this extra enum?

dmoj/judge.py
```diff
@@ -459,69 +491,132 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+            tests: List[TopLevelCase], should_run: Callable[[TopLevelCase, Set[int]], bool]
+        ) -> Generator[Tuple[IPC, tuple], None, Tuple[Set[int], GradeManyResult]]:
+            failed_tests: Set[int] = set()
+            test_iter = iter(tests)
+            for test in test_iter:
+                if not should_run(test, failed_tests):
+                    failed = True
+                    yield from skip_toplevel(test)
+                elif isinstance(test, Batch):
+                    yield IPC.BATCH_BEGIN, (test.batch,)
```
Xyene, line 556: Can we move these yields into `grade_batched_cases`? It's a little weird that the case results come from within it, but the batch begin/end markers don't.

dmoj/judge.py
```diff
@@ -459,69 +491,132 @@ def _grade_cases(self) -> Generator[Tuple[IPC, tuple], None, None]:
...
+
+                    batch_verdict = yield from grade_batched_cases(test.batched_cases)
+                    failed = batch_verdict != Result.AC
+
+                    yield IPC.BATCH_END, (test.batch,)
                 else:
-                    result = self.grader.grade(case)
+                    result = grade(test)
+                    failed = result.result_flag & Result.WA
+                    yield IPC.RESULT, (None, test.position, result)
```
Xyene, line 565: Can we move this yield into `grade`?

PR #1077, opened 2022-11-10
executors/RUST: build rust into shared directory

Riolku:
Because our list of dependencies grew, so did our compile times. This is a much bigger deal because rust compiles dependencies even if they are unused. See [this submission](https://dmoj.ca/submission/5018534) to Hello, World from myself.

This PR adds an option to make Rust compile into a shared directory, which shares dependencies.

From the perspective of `cargo`, it should seem like we are the same package, changing our source code every time a submission comes in.

### DIFF
--- a/dmoj/executors/RUST.py
+++ b/dmoj/executors/RUST.py
@@ -1,3 +1,4 @@
+import fcntl
 import os
 
 from dmoj.cptbox.filesystem_policies import ExactFile, RecursiveDir
@@ -56,15 +57,44 @@ class Executor(CompiledExecutor):
         with open(self._file('Cargo.toml'), 'wb') as f:
             f.write(CARGO_TOML)
 
+    shared_target = None
+
+    @classmethod
+    def get_shared_target(cls):
+        if cls.shared_target is not None:
+            return cls.shared_target
+
+        cargo_dir = os.path.expanduser('~/.cargo')
+        collisions = 0
+        while True:
+            maybe_target = os.path.join(cargo_dir, f'dmoj-shared-target-{collisions}')
+            try:
+                os.mkdir(maybe_target, mode=0o775)
+            except FileExistsError:
+                pass
+
+            dirfd = os.open(maybe_target, os.O_RDONLY | os.O_DIRECTORY | os.O_CLOEXEC)
+            try:
+                fcntl.flock(dirfd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                # dirfd is leaked on purpose.
+            except BlockingIOError:
+                # Another judge is using this.
+                os.close(dirfd)
+                collisions += 1
+            else:
+                cls.shared_target = maybe_target
+                # We intentionally don't clean this directory up at any point, since we can re-use it on restart.
+                return cls.shared_target
+
     @classmethod
     def get_versionable_commands(cls):
         return [('rustc', os.path.join(os.path.dirname(cls.get_command()), 'rustc'))]
 
     def get_compile_args(self):
-        args = [self.get_command(), 'build', '--release']
+        args = [self.get_command(), 'build', '--release', '--target-dir', self.get_shared_target()]
         if bool_env('DMOJ_CARGO_OFFLINE'):
             args += ['--offline']
         return args
 
     def get_compiled_file(self):
-        return self._file('target', 'release', 'user_submission')
+        return os.path.join(self.get_shared_target(), 'release/user_submission')


Riolku:
With this PR, only the first submission of every judge lifetime will compile dependencies for rust.

Riolku:
Of particular concern is generators or checkers written in rust, which would cause a significant overhead to any problems with them.

dmoj/executors/RUST.py
```diff
@@ -62,9 +62,14 @@ def get_versionable_commands(cls):
...
+        if bool_env('DMOJ_CARGO_SHARED_TARGET'):
+            args += ['--target-dir', os.path.expanduser('~/.cargo/dmoj-shared-target')]
         if bool_env('DMOJ_CARGO_OFFLINE'):
             args += ['--offline']
         return args
 
     def get_compiled_file(self):
-        return self._file('target', 'release', 'user_submission')
+        if bool_env('DMOJ_CARGO_SHARED_TARGET'):
+            return os.path.expanduser('~/.cargo/dmoj-shared-target/release/user_submission')
```
Xyene, line 73: We should embed some sort of UUID here (and maybe ensure it doesn't already exist at judge startup). Running multiple judge instances from the same machine is a supported configuration that'd break as-is (though prod would be fine because of Docker).

dmoj/executors/RUST.py
```diff
@@ -62,9 +62,14 @@ def get_versionable_commands(cls):
 
     def get_compile_args(self):
         args = [self.get_command(), 'build', '--release']
+        if bool_env('DMOJ_CARGO_SHARED_TARGET'):
```
Xyene, line 65: I'm fine making this the default and not having this option.

dmoj/executors/RUST.py
Riolku: Done.

dmoj/executors/RUST.py
```diff
@@ -56,15 +59,44 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
+                maybe_target = cargo_dir + f'/dmoj-shared-target-{id_hash}-{collisions}'
+                try:
+                    os.mkdir(maybe_target, mode=0o775)
+                except FileExistsError:
+                    pass
+                try:
+                    # This is 'leaked' on purpose.
+                    dirfd = os.open(maybe_target, os.O_RDONLY | os.O_CLOEXEC)
+                    fcntl.flock(dirfd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                except BlockingIOError:
```
Xyene, line 83: Missing increment of collisions (also failing mypy elsewhere)

dmoj/executors/RUST.py
```diff
@@ -56,15 +59,44 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
+    shared_target = None
+
+    @classmethod
+    def get_shared_target(cls):
+        if cls.shared_target is None:
+            cargo_dir = os.path.expanduser('~/.cargo')
+            collisions = 0
+            judge_id = env.get('id', 'anonymous')
+            hasher = hashlib.sha256()
+            hasher.update(judge_id.encode('utf-8'))
```
Xyene, line 71: After thinking about this more, by doing the flock thing we don't even need to include the hash in the name, just the collision count.

dmoj/executors/RUST.py
Riolku: Can you merge #1078 to fix mypy?

Riolku: Also thanks, fixed.

dmoj/executors/RUST.py
Riolku: Done.

dmoj/executors/RUST.py
```diff
@@ -56,15 +57,40 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
+                try:
+                    os.mkdir(maybe_target, mode=0o775)
+                except FileExistsError:
+                    pass
+                try:
+                    # This is 'leaked' on purpose.
+                    dirfd = os.open(maybe_target, os.O_RDONLY | os.O_CLOEXEC)
+                    fcntl.flock(dirfd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                except BlockingIOError:
+                    collisions += 1
```
Xyene, line 78: We shouldn't leak dirfd in this branch.

dmoj/executors/RUST.py
```diff
@@ -56,15 +57,40 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
         with open(self._file('Cargo.toml'), 'wb') as f:
             f.write(CARGO_TOML)
 
+    shared_target = None
+
+    @classmethod
+    def get_shared_target(cls):
+        if cls.shared_target is None:
```
Xyene, line 64: If we invert this condition we can reduce nesting by one level.

dmoj/executors/RUST.py
Riolku: Agreed, and honestly, we shouldn't have it in the `try` at all.

dmoj/executors/RUST.py
Riolku: Updated.

dmoj/executors/RUST.py
```diff
@@ -56,15 +57,42 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
+            return cls.shared_target
+
+        cargo_dir = os.path.expanduser('~/.cargo')
+        collisions = 0
+        while True:
+            maybe_target = cargo_dir + f'/dmoj-shared-target-{collisions}'
+            try:
+                os.mkdir(maybe_target, mode=0o775)
+            except FileExistsError:
+                pass
```
quantum5, line 74: Is the intention here to never clean up the files in `.cargo/dmoj-shared-target-...`?

Xyene: Yes, and to not collide with other judge processes when running outside Docker (hence the flocks). I guess this could use a comment though.

Riolku: Commented.

dmoj/executors/RUST.py
```diff
@@ -56,15 +57,44 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
+            return cls.shared_target
+
+        cargo_dir = os.path.expanduser('~/.cargo')
+        collisions = 0
+        while True:
+            maybe_target = cargo_dir + f'/dmoj-shared-target-{collisions}'
+            try:
+                os.mkdir(maybe_target, mode=0o775)
+            except FileExistsError:
+                # Another judge is using this.
```
Xyene, line 74: This is not necessarily true after a restart.

Riolku: Wrong place, this comment should be under the `blocking` error... one sec.

Riolku: Updated.

dmoj/executors/RUST.py
```diff
@@ -56,15 +57,44 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
+        cargo_dir = os.path.expanduser('~/.cargo')
+        collisions = 0
+        while True:
+            maybe_target = cargo_dir + f'/dmoj-shared-target-{collisions}'
+            try:
+                os.mkdir(maybe_target, mode=0o775)
+            except FileExistsError:
+                pass
+
+            dirfd = os.open(maybe_target, os.O_RDONLY | os.O_CLOEXEC)
```
Xyene, line 76: I think this should work and provide a little more validation 

```suggestion
            dirfd = os.open(maybe_target, os.O_PATH | os.O_DIRECTORY | os.O_CLOEXEC)
```

otherwise LGTM

Riolku: `O_PATH` doesn't work, but I added `O_DIRECTORY`.

If we use `O_PATH`, `flock` fails with EBADF.

dmoj/executors/RUST.py
```diff
@@ -56,15 +57,44 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
+
+    @classmethod
+    def get_shared_target(cls):
+        if cls.shared_target is not None:
+            return cls.shared_target
+
+        cargo_dir = os.path.expanduser('~/.cargo')
+        collisions = 0
+        while True:
+            maybe_target = cargo_dir + f'/dmoj-shared-target-{collisions}'
```
quantum5, line 70: By the way can we use proper path concatenation?

dmoj/executors/RUST.py
```diff
@@ -56,15 +57,44 @@ def create_files(self, problem_id, source_code, *args, **kwargs):
...
     def get_compile_args(self):
-        args = [self.get_command(), 'build', '--release']
+        args = [self.get_command(), 'build', '--release', '--target-dir', self.get_shared_target()]
         if bool_env('DMOJ_CARGO_OFFLINE'):
             args += ['--offline']
         return args
 
     def get_compiled_file(self):
-        return self._file('target', 'release', 'user_submission')
+        return self.get_shared_target() + '/release/user_submission'
```
quantum5, line 100: Proper path concatenation?

dmoj/executors/RUST.py
Riolku: Do you mean `os.path.join`, or `pathlib`, or something else?

Riolku: Using `os.path.join`.

Xyene (approved):
LGTM, thanks!

PR #1081, opened 2022-11-15
executors/RUST: Use per-executor shared directory

Riolku:
After #1077, in checkers and interactors, if both the submission and checker/interactor are bridged, they will trample eachother's target directory.

We can fix this by making shared target directories per-executor.

When we use a bridged Rust checker with a Rust submission, the first case works just fine, but the second one runs the checker binary (which probably RTEs).


Xyene (approved):
LGTM, thanks

PR #1091, opened 2022-12-27
executors/BF: optimize BF using LLVM

int-y1:
There are a few parts to this BF optimizing compiler: simple loop pass, translation into LLVM, `opt`, `llc -O2`, `ld`.

### Simple loop pass (Python)

Optimizes simple loops. `opt` can optimize these loops too, but it takes a long time to do so. If this step is skipped, the BF program [1162081](https://dmoj.ca/src/1162081/raw) takes 67s on `opt` and gets compiler TLE (limit is 10s).

Also checks for unmatched brackets. Later steps don't need to check for unmatched brackets.

### Translation into LLVM

A few notable things:

* `mmap` rounds up to the nearest 4096, so I put `memory: 4` in every `test.yml`.
* The pointer looks like `%p`, and is maintained using `getelementptr` and `phi`.
* I used `*char_unlocked`, because for large I/O, it's much faster than `*char`.

### `opt`

Optimizes LLVM using the passes in `OPT_PASSES`. The performance gain is very tiny (~5%), and I suspect it's because `llc -O2` does its own optimizations. I didn't use `opt -O2` for 2 reasons: possible performance regressions in the future, and similar performance to `OPT_PASSES`.

### `llc -O2`, `ld`

These are the LLC executor's defaults. I didn't use `llc -O0`, because in every test, `O0` produced slower programs than `O2`.

## Benchmarks

These benchmarks were done using my slow laptop. For each number, I ran 5-9 trials and took the median. The unit is milliseconds.

Prod currently uses `gcc -O0`. The goal to beat is `gcc -O2` + `putchar_unlocked`.

| Program | method of compilation | `opt` step | total compilation | runtime |
| - | - | - | - | - |
| [mandelbrot.b](https://rosettacode.org/wiki/Mandelbrot_set#Brainf***) (11K inst, 6K output) | `gcc -O0` | - | 841 | 28746 |
| mandelbrot.b | `gcc -O2` | - | 2480 | 1648 |
| mandelbrot.b | `clang -O2` | - | 4925 | 1790 |
| mandelbrot.b | this PR | 706 | 1237 | 1627 |
| [1162081](https://dmoj.ca/src/1162081/raw) (46K inst, 16M output) | `gcc -O0` | - | 5371 | 30913 |
| 1162081 | `gcc -O2` | - | 12579 | 1231 |
| 1162081 | `gcc -O2` + `putchar_unlocked` | - | 23962 | 1010 |
| 1162081 | `clang -O2` | - | >120000 💀 | ? |
| 1162081 | this PR | 1433 | 2892 | 1132 |
| [230309](https://dmoj.ca/src/230309/raw) (240 inst, 16M output) | `gcc -O0` | - | 325 | 1135 |
| 230309 | `gcc -O2` | - | 315 | 297 |
| 230309 | `gcc -O2` + `putchar_unlocked` | - | 352 | 195 |
| 230309 | `clang -O2` + `putchar_unlocked` | - | 304 | 211 |
| 230309 | this PR | 83 | 184 | 206 |


kiritofeng:
ok to test

int-y1:
[Prod tests](https://gist.github.com/Xyene/0d6babb0db264d784593e60af3b3a0e4). Methodology:
* Rejudge everything with LLVM to get `new_*` columns.
* Rejudge everything with `gcc -O0` to get `old_*` columns. This ensures the old data is recent.

My notes:
* Compiler TLE is so minor that I don't consider it an issue.
* There are lots of TLE->AC, which is good. Some took longer to RTE/TLE, but that's fine. Some ACs took longer (e.g. 2968660, 1359681), but there's [evidence](https://gist.github.com/Xyene/a490c41a2b3417b70877c56018a373cb) that it's noise.
* Memory use was roughly the same, because both executors use libc.

I didn't see any issues with verdict change, execution time, or memory use.

testsuite/helloworld/tests/bf_rte_low/test.yml
```diff
@@ -0,0 +1,5 @@
+language: BF
+time: 1
+memory: 4
```
Xyene, line 3: `mmap` rounding to page size is not documented and is broken in some cases, so let's just specify a reasonable limit here.

dmoj/executors/BF.py
```diff
@@ -1,80 +1,215 @@
...
+        elif b in b',.':
+            bf_inst.append(chr(b))
+            simple = False
+        elif b == b'['[0]:
+            bf_inst.append('[')
+            bf_stack.append(len(bf_inst))
+            simple = True
+        elif b == b']'[0]:
+            if not bf_stack:
+                raise CompileError(b'Unmatched brackets\n')
```
Xyene, line 89:
```suggestion
                raise CompileError(b'Unmatched brackets')
```

dmoj/executors/BF.py
```diff
@@ -1,80 +1,215 @@
...
+            i = bf_stack.pop()
+            loop = simple_loop(bf_inst[i:]) if simple else None
+            if loop is None:
+                bf_inst.append(']')
+            else:
+                while len(bf_inst) >= i:
+                    bf_inst.pop()
```
quantum5, line 96: Why not just `bf_inst = bf_inst[:i+1]` or something like that?

testsuite/helloworld/tests/bf_rte_low/test.yml
int-y1: Upgraded 4 to 64.

dmoj/executors/BF.py
```diff
@@ -1,80 +1,215 @@
...
-                if open_brackets == 0:
-                    return True
-                else:
-                    open_brackets -= 1
-        return open_brackets != 0
+    # Do both opt and llc.
+    def assemble(self) -> Tuple[bytes, List[str]]:
+        assert self._code is not None
+        # TODO: find an alternative to get opt_path
+        opt_path = self.get_as_path().rstrip(self.as_name) + 'opt'
```
quantum5, line 196: I am horrified by this since `rstrip` removes any character that appears in `self.as_name` until it sees one that doesn't. Can you at least do
```suggestion
        opt_path = re.sub(f'{re.escape(self.as_name)}$', 'opt', self.get_as_path())
```

dmoj/executors/BF.py
int-y1: This loop is O(popped items). `bf_inst = bf_inst[:i+1]` is O(i) and will make compilation O(N^2).

quantum5: Have you tried `bf_inst[i+1:] = []` then?

dmoj/executors/BF.py
int-y1: Done.

dmoj/executors/BF.py
int-y1: Done. `bf_inst[i - 1 :] = [loop]` runs fast.

dmoj/executors/BF.py
```diff
@@ -1,80 +1,215 @@
...
-    ord('['): b'while(*p){',
-    ord(']'): b'}',
-}
 
+# Helper for simple loops. Consider the loop "[>>++>-<<<-]".
+# This helper translates it into {2: 2, 3: -1}. Later, the LLVM looks like this:
+#   if (p[0] != 0) { p[2] += p[0] * 2; p[3] += p[0] * -1; p[0] = 0; }
+# The if-guard is needed so that programs like "[-<+>]" don't segfault.
+def simple_loop(bf_inst: List[Union[str, Dict[int, int]]]) -> Optional[Dict[int, int]]:
+    loop: Dict[int, int] = {}
```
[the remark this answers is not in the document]
int-y1: Done.

Xyene (approved):
LGTM, thanks for working on this!

PR #1092, opened 2022-12-28
Make LLC executor work on non-amd64 platforms

quantum5:
This appears to work, but we need to get this merged before the runtimes will successfully build, which will enable the tests to pass.


Xyene (approved):
LGTM, thanks!

PR #1093, opened 2022-12-31
Add input validation command

kiritofeng:
Add to `init.yml`

```yaml
validator:
  source: validate.cpp
  language: CPP14
  time: 2
  memory: 65536
```

### DIFF
--- a/dmoj/commands/validate.py
+++ b/dmoj/commands/validate.py
@@ -1,0 +1,176 @@
+import os
+import shlex
+import subprocess
+from itertools import groupby
+from operator import itemgetter
+from typing import List, Optional, Tuple
+
+
+from dmoj import executors
+from dmoj.commands.base_command import Command
+from dmoj.contrib import contrib_modules
+from dmoj.error import CompileError, InvalidCommandException, OutputLimitExceeded
+from dmoj.judgeenv import env, get_problem_root, get_supported_problems
+from dmoj.problem import BatchedTestCase, Problem, ProblemConfig, ProblemDataManager, TestCase
+from dmoj.result import Result
+from dmoj.utils.ansi import print_ansi
+from dmoj.utils.helper_files import compile_with_auxiliary_files
+from dmoj.utils.unicode import utf8text
+
+
+all_executors = executors.executors
+
+
+class ValidateCommand(Command):
+    name = 'validate'
+    help = 'Validates input for problems.'
+
+    def _populate_parser(self) -> None:
+        self.arg_parser.add_argument('problem_ids', nargs='+', help='ids of problems to validate input')
+
+    def execute(self, line: str) -> int:
+        args = self.arg_parser.parse_args(line)
+
+        problem_ids = args.problem_ids
+        supported_problems = set(get_supported_problems())
+
+        unknown_problems = ', '.join(
+            f"'{problem_id}'" for problem_id in problem_ids if problem_id not in supported_problems
+        )
+        if unknown_problems:
+            raise InvalidCommandException(f'unknown problem(s) {unknown_problems}')
+
+        total_fails = 0
+        for problem_id in problem_ids:
+            if not self.validate_problem(problem_id):
+                print_ansi(f'Problem #ansi[{problem_id}](cyan|bold) #ansi[failed validation](red|bold).')
+                total_fails += 1
+            else:
+                print_ansi(f'Problem #ansi[{problem_id}](cyan|bold) passed with flying colours.')
+            print()
+
+        print()
+        print('Input validation complete.')
+        if total_fails:
+            print_ansi(f'#ansi[A total of {total_fails} problem(s) have invalid input.](red|bold)')
+        else:
+            print_ansi('#ansi[All problems validated.](green|bold)')
+
+        return total_fails
+
+    def validate_problem(self, problem_id: str) -> bool:
+        print_ansi(f'Validating problem #ansi[{problem_id}](cyan|bold)...')
+
+        problem_root = get_problem_root(problem_id)
+        if problem_root is None:
+            print_ansi(f'\t#ansi[Skipped](magenta|bold) - Problem [{problem_id}](cyan|bold) not found.')
+            return True
+
+        config = ProblemConfig(ProblemDataManager(problem_root))
+        if not config.validator:
+            print_ansi('\t#ansi[Skipped](magenta|bold) - No validator found.')
+            return True
+
+        validator_config = config['validator']
+        language = validator_config['language']
+        if language not in all_executors:
+            print_ansi('\t\t#ansi[Skipped](magenta|bold) - Language not supported.')
+            return True
+
+        time_limit = validator_config.time_limit or env.validator_time_limit
+        memory_limit = validator_config.memory_limit or env.validator_memory_limit
+        compiler_time_limit = validator_config.get('compiler_time_limit', env.validator_compiler_time_limit)
+        read_feedback_from = validator_config.get('read_feedback_from', 'stderr')
+        if read_feedback_from not in ('stdout', 'stderr'):
+            print_ansi('\t\t#ansi[Failed](red|bold) - Feedback option should be (stdout, stderr).')
+            return False
+
+        if isinstance(validator_config.source, str):
+            filenames = [validator_config.source]
+        elif isinstance(validator_config.source.unwrap(), list):
+            filenames = list(validator_config.source.unwrap())
+        else:
+            print_ansi('\t#ansi[Failed](red|bold) - No validator found.')
+            return False
+
+        filenames = [os.path.abspath(os.path.join(problem_root, name)) for name in filenames]
+        try:
+            executor = compile_with_auxiliary_files(filenames, lang=language, compiler_time_limit=compiler_time_limit)
+        except CompileError as compilation_error:
+            print_ansi('#ansi[Failed compiling validator!](red|bold)')
+            print(compilation_error.message.rstrip())
+            return False
+
+        problem = Problem(problem_id, time_limit, memory_limit, {})
+
+        flattened_cases: List[Tuple[Optional[int], TestCase]] = []
+        batch_number = 0
+        for case in problem.cases():
+            if isinstance(case, BatchedTestCase):
+                batch_number += 1
+                for batched_case in case.batched_cases:
+                    assert isinstance(batched_case, TestCase)
+                    flattened_cases.append((batch_number, batched_case))
+            else:
+                assert isinstance(case, TestCase)
+                flattened_cases.append((None, case))
+
+        contrib_type = validator_config.get('type', 'default')
+        if contrib_type not in contrib_modules:
+            print_ansi(f'#ansi[{contrib_type} is not a valid contrib module!](red|bold)')
+            return False
+
+        args_format_string = (
+            validator_config.args_format_string
+            or contrib_modules[contrib_type].ContribModule.get_validator_args_format_string()
+        )
+
+        case_number = 0
+        ok = True
+        for batch_number, cases in groupby(flattened_cases, key=itemgetter(0)):
+            if batch_number:
+                print_ansi(f'#ansi[Batch #{batch_number}](yellow|bold)')
+            for _, case in cases:
+                case_number += 1
+
+                result = Result(case)
+                input = case.input_data()
+
+                validator_args = shlex.split(args_format_string.format(batch_no=case.batch, case_no=case.position))
+                process = executor.launch(
+                    *validator_args,
+                    time=time_limit,
+                    memory=memory_limit,
+                    stdin=subprocess.PIPE,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    wall_time=case.config.wall_time_factor * time_limit,
+                )
+                try:
+                    proc_output, proc_error = process.communicate(
+                        input, outlimit=case.config.output_limit_length, errlimit=1048576
+                    )
+                except OutputLimitExceeded:
+                    proc_error = b''
+                    process.kill()
+                finally:
+                    process.wait()
+
+                executor.populate_result(proc_error, result, process)
+                feedback = (
+                    utf8text({'stdout': proc_output, 'stderr': proc_error}[read_feedback_from].rstrip())
+                    or result.feedback
+                )
+                code = result.readable_codes()[0]
+                code_colour = Result.COLORS_BYID[code]
+                if code == 'AC':
+                    code = 'OK'  # this is less confusing
+                colored_code = f'#ansi[{code}]({code_colour}|bold)'
+                colored_feedback = f'(#ansi[{utf8text(feedback)}](|underline))' if feedback else ''
+                case_padding = '  ' if batch_number is not None else ''
+                print_ansi(f'{case_padding}Test case {case_number:2d} {colored_code:3s} {colored_feedback}')
+
+                if result.result_flag:
+                    ok = False
+
+        return ok
--- a/dmoj/contrib/coci.py
+++ b/dmoj/contrib/coci.py
@@ -14,6 +14,10 @@ class ContribModule(TestlibContribModule):
         return '{input_file} {answer_file}'
 
     @classmethod
+    def get_validator_args_format_string(cls):
+        raise NotImplementedError
+
+    @classmethod
     def parse_return_code(cls, proc, executor, point_value, time_limit, memory_limit, feedback, name, stderr):
         if proc.returncode == cls.PARTIAL:
             match = cls.repartial.search(stderr)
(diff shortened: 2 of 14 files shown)


kiritofeng:
Sample output:
```
dmoj> validate echo
Validating problem echo...
Test case  1 AC (OK)
Batch #1
  Test case  2 AC (OK)
  Test case  3 IR (AssertionError) {WA}
Batch #2
  Test case  4 AC (OK)
  Test case  5 IR (AssertionError) {WA}
Problem echo failed.
```

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+        if not config or 'validator' not in config:
+            print_ansi('\t#ansi[Skipped](magenta|bold) - No validator found')
+            return 0
+
+        validator_config = config['validator']
+        language = validator_config['language']
+        if language not in all_executors:
+            print_ansi('\t\t#ansi[Skipped](magenta|bold) - Language not supported')
+            return 0
+        time_limit = validator_config.get('time', env.generator_time_limit)
```
Xyene, line 74: Does it make sense to have this be configurable? Either way I think we should have a `validator_time_limit` field in `env` rather than reuse the generator limit, even if we do end up setting them to the same values.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+        memory_limit = validator_config.get('memory', env.generator_memory_limit)
+        compiler_time_limit = validator_config.get('compiler_time_limit', env.generator_compiler_time_limit)
+        read_feedback_from = validator_config.get('feedback', 'stderr')
```
Xyene, line 77: Let's give this field a more detailed name.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+
+        if read_feedback_from not in ('stdout', 'stderr'):
+            print_ansi('\t\t#ansi[Skipped](magenta|bold) - Feedback option not supported')
```
Xyene, line 80: Embed the valid options in the error message.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+            return 0
```
Xyene, line 81: This feels like it should be an error?

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+            executor = compile_with_auxiliary_files(filenames, [], language, compiler_time_limit)
+        except CompileError as compilation_error:
+            error_msg = compilation_error.message or 'compiler exited abnormally'
+            print_ansi('#ansi[Failed compiling validator!](red|bold)')
+            print(error_msg.rstrip())
+            return 1
+
+        problem = Problem(problem_id, time_limit, memory_limit, {})
+
+        grader_class = type('Grader', (problem.grader_class,), {'_generate_binary': lambda *_: None})  # don't compile
```
Xyene, line 102: This deserves a comment about what we're doing here. Why are we using our own compilation routine instead of the grader's?

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+            filenames = list(validator_config.source.unwrap())
+        else:
+            print_ansi('\t#ansi[Skipped](magenta|bold) - No validator found')
+            return 0
+
+        filenames = [os.path.abspath(os.path.join(problem_root, name)) for name in filenames]
+        try:
+            executor = compile_with_auxiliary_files(filenames, [], language, compiler_time_limit)
+        except CompileError as compilation_error:
+            error_msg = compilation_error.message or 'compiler exited abnormally'
```
Xyene, line 95: Rather than duplicating this message here, what if we made it the default when creating a `CompileError`?

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+
+                result = Result(case)
+                input = case.input_data()
+
+                validator_args = shlex.split(args_format_string.format(batch_no=case.batch, case_no=case.position))
+                process = executor.launch(
+                    *validator_args,
+                    time=time_limit,
+                    memory=memory_limit,
+                    symlinks=case.config.symlinks,
```
Xyene, line 141: I don't think we need to be passing symlinks here.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,173 @@
...
+                )
+
+                code = result.readable_codes()[0]
+                colored_code = f'#ansi[{code}]({Result.COLORS_BYID[code]}|bold)'
+                colored_feedback = f'(#ansi[{utf8text(feedback)}](|underline))' if feedback else ''
+                case_padding = '  ' if batch_number is not None else ''
+                print_ansi(f'{case_padding}Test case {case_number:2d} {colored_code:3s} {colored_feedback}')
+
+                if result.result_flag:
+                    fail = 1
```
Xyene, line 171: This feels like it should be a boolean.

dmoj/commands/validate.py
kiritofeng: Graders don't easily allow for multi-file compilation (but we need the grader object to resolve the testcases, and specifically their order) also there's no guarantee the grader hasn't overwritten the compilation routine to do something weird, so I just thought it would be easier to forcibly turn it off.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,182 @@
...
+
+        if not config or 'validator' not in config:
+            print_ansi('\t#ansi[Skipped](magenta|bold) - No validator found')
+            return True
+
+        validator_config = config['validator']
+        language = validator_config['language']
+        if language not in all_executors:
+            print_ansi('\t\t#ansi[Skipped](magenta|bold) - Language not supported')
+            return True
```
Riolku, line 77: Add line after.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,182 @@
...
+        time_limit = validator_config.get('time', env.validator_time_limit)
+        memory_limit = validator_config.get('memory', env.validator_memory_limit)
+        compiler_time_limit = validator_config.get('compiler_time_limit', env.validator_compiler_time_limit)
+        read_feedback_from = validator_config.get('read_feedback_from', 'stderr')
+
```
Riolku, line 82: Delete line.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,182 @@
...
+            print_ansi('\t\t#ansi[Failed](red|bold) - Feedback option should be (stdout, stderr)')
+            return False
+
+        if isinstance(validator_config.source, str):
+            filenames = [validator_config.source]
+        elif isinstance(validator_config.source.unwrap(), list):
+            filenames = list(validator_config.source.unwrap())
+        else:
+            print_ansi('\t#ansi[Skipped](magenta|bold) - No validator found')
+            return True
```
Riolku, line 93: If a validator config exists but no sources are found, maybe this isn't "skip"? Maybe this should be a failure?

dmoj/commands/validate.py
```diff
@@ -0,0 +1,182 @@
...
+
+        filenames = [os.path.abspath(os.path.join(problem_root, name)) for name in filenames]
+        try:
+            executor = compile_with_auxiliary_files(filenames, [], language, compiler_time_limit)
```
Riolku, line 97: Does python have good support for specifying what the empty list represents?

dmoj/commands/validate.py
```diff
@@ -0,0 +1,182 @@
...
+                        input, outlimit=case.config.output_limit_length, errlimit=1048576
+                    )
+                except OutputLimitExceeded:
+                    proc_error = b''
+                    process.kill()
+                finally:
+                    process.wait()
+
+                executor.populate_result(proc_error, result, process)
+
```
Riolku, line 167: Delete line.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,182 @@
...
+            print(error_msg.rstrip())
+            return False
+
+        problem = Problem(problem_id, time_limit, memory_limit, {})
+
+        # We need to override the compilation, because they don't allow for
+        # easy multi-file compilation (e.g. testlib.h) and we also don't know
+        # if the grader has overwritten the compilation routine.
+        # The downside is validating signature graded problems requires
+        # manually setting up the entry point by hand.
```
Riolku, line 110: We have to override the compilation? But the compilation should already be handled, right? Wait what is happening

dmoj/contrib/coci.py
```diff
@@ -1,27 +1,47 @@
...
     @classmethod
-    def parse_return_code(cls, proc, executor, point_value, time_limit, memory_limit, feedback, name, stderr):
+    def get_validator_args_format_string(cls) -> str:
+        raise NotImplementedError
+
+    @classmethod
+    def parse_return_code(
+        cls,
+        proc: 'TracedPopen',
+        executor: 'BaseExecutor',
```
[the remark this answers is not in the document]
quantum5: You definitely still need to import the name if using strings, but you should be able to use the identifier directly here.

dmoj/contrib/default.py
```diff
@@ -1,23 +1,72 @@
...
 from dmoj.result import CheckerResult
 from dmoj.utils.helper_files import parse_helper_file_error
 
+if TYPE_CHECKING:
+    from dmoj.cptbox import TracedPopen
+    from dmoj.executors.base_executor import BaseExecutor
+
 
-class ContribModule:
+class BaseContribModule:
```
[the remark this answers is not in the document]
Riolku: (Like, create that file).

dmoj/commands/validate.py
```diff
@@ -0,0 +1,182 @@
...
+        else:
+            print_ansi('#ansi[All problems validated.](green|bold)')
+
+        return total_fails
+
+    def validate_problem(self, problem_id: str) -> bool:
+        print_ansi(f'Validating problem #ansi[{problem_id}](cyan|bold)...')
+
+        problem_root = get_problem_root(problem_id)
+        assert problem_root is not None
```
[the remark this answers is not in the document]
kiritofeng: Ah this is an error stemming from me forgetting how the code works

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+
+        problem = Problem(problem_id, time_limit, memory_limit, {})
+        # We need to override the grader's own compilation routine,
+        # because they don't allow for easy multi-file compilation
+        # (e.g. testlib.h) and we also don't know if the grader has already
+        # overwritten the compilation routine (e.g. signature grading, custom grader).
+        # The downside is validating signature graded problems requires
+        # manually setting up the entry point by hand.
+        grader_class = type('Grader', (problem.grader_class,), {'_generate_binary': lambda *_: None})
+        grader = grader_class(self.judge, problem, language, b'')
```
Riolku, line 111: Can I get a walkthrough of why the code works like this? The way I imagine validators working is we get the input data, and run the executor directly on the input. I don't feel like using a Grader is important here... what logic are we making use of? Just the case fetching? Can we extract that some other way?

As is, I'm worried that this is a hack that will bite us later.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+                executor.populate_result(proc_error, result, process)
+                feedback = (
+                    utf8text({'stdout': proc_output, 'stderr': proc_error}[read_feedback_from].rstrip())
+                    or result.feedback
+                )
+                code = result.readable_codes()[0]
+                colored_code = f'#ansi[{code}]({Result.COLORS_BYID[code]}|bold)'
+                colored_feedback = f'(#ansi[{utf8text(feedback)}](|underline))' if feedback else ''
+                case_padding = '  ' if batch_number is not None else ''
+                print_ansi(f'{case_padding}Test case {case_number:2d} {colored_code:3s} {colored_feedback}')
```
Riolku, line 174: Using `Result` might be overkill here. Thoughts?

Riolku: I guess re-handling all the codes is kind of a pain, nevermind.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+    def execute(self, line: str) -> int:
+        args = self.arg_parser.parse_args(line)
+
+        problem_ids = args.problem_ids
+        supported_problems = set(get_supported_problems())
+
+        unknown_problems = ', '.join(f"'{i}'" for i in problem_ids if i not in supported_problems)
+        if unknown_problems:
+            raise InvalidCommandException(f'unknown problem(s) {unknown_problems}')
+        total_fails = 0
```
Xyene, line 40:
```suggestion

        total_fails = 0
```

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+                print_ansi(f'Problem #ansi[{problem_id}](cyan|bold) #ansi[failed](red|bold).')
+                total_fails += 1
+            else:
+                print_ansi(f'Problem #ansi[{problem_id}](cyan|bold) passed with flying colours.')
+            print()
+
+        print()
+        print('Test complete.')
+        if total_fails:
+            print_ansi(f'#ansi[A total of {total_fails} problem(s) have invalid input](red|bold)')
```
Xyene, line 52:
```suggestion
            print_ansi(f'#ansi[A total of {total_fails} problem(s) have invalid input.](red|bold)')
```

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def validate_problem(self, problem_id: str) -> bool:
+        print_ansi(f'Validating problem #ansi[{problem_id}](cyan|bold)...')
+
+        problem_root = get_problem_root(problem_id)
+        if problem_root is None:
+            print_ansi(f'\t#ansi[Skipped](magenta|bold) - Problem [{problem_id}](cyan|bold) not found')
+            return True
+
+        assert problem_root is not None
```
Xyene, line 66: This shouldn't be necessary to appease mypy, since we guard with `if problem_root is None: ... return` above.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+        config = ProblemConfig(ProblemDataManager(problem_root))
+        if not config or 'validator' not in config:
```
Xyene, line 68: I think since this is a `ProblemConfig`, `if not config.validator` should work here, without needing the `not config` guard.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+            print_ansi('\t#ansi[Skipped](magenta|bold) - No validator found')
+            return True
+
+        validator_config = config['validator']
+        language = validator_config['language']
+        if language not in all_executors:
+            print_ansi('\t\t#ansi[Skipped](magenta|bold) - Language not supported')
+            return True
+
+        time_limit = validator_config.get('time', env.validator_time_limit)
```
Xyene, line 78: `validator_config.time or env.validator_time_limit` etc.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+        memory_limit = validator_config.get('memory', env.validator_memory_limit)
```
Xyene, line 79: Let's suffix limits with `_limit`.

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+            print_ansi('\t#ansi[Failed](red|bold) - No validator found')
+            return False
+
+        filenames = [os.path.abspath(os.path.join(problem_root, name)) for name in filenames]
+        try:
+            executor = compile_with_auxiliary_files(filenames, lang=language, compiler_time_limit=compiler_time_limit)
+        except CompileError as compilation_error:
+            error_msg = compilation_error.message
+            print_ansi('#ansi[Failed compiling validator!](red|bold)')
+            print(error_msg.rstrip())
```
Xyene, line 100:
```suggestion
            print(compilation_error.message.rstrip())
```

dmoj/commands/validate.py
```diff
@@ -0,0 +1,179 @@
...
+            filenames = list(validator_config.source.unwrap())
+        else:
+            print_ansi('\t#ansi[Failed](red|bold) - No validator found')
+            return False
+
+        filenames = [os.path.abspath(os.path.join(problem_root, name)) for name in filenames]
+        try:
+            executor = compile_with_auxiliary_files(filenames, lang=language, compiler_time_limit=compiler_time_limit)
+        except CompileError as compilation_error:
+            error_msg = compilation_error.message
```
Xyene, line 98:
```suggestion
```

Xyene (approved):
LGTM

Riolku (approved):

Riolku:
Yay for sanity - thanks for fixing it up.

quantum5 (approved):

PR #1096, opened 2023-01-02
isolate: globally allow epoll syscalls

Xyene:
It's used in enough runtimes, and apparently both PyPy and Python 3 use it sometimes (but weren't allowed to do so until now).


quantum5 (approved):
`epoll` seem safe enough.

PR #1102, opened 2023-01-23
Remove deprecated `problem_storage_root`

kiritofeng:


kiritofeng:
Decided to leave the test suite as-is, because I realize that using globs isn't as nice as I thought it'd be.

.docker/entry
```diff
@@ -11,7 +11,7 @@ pip3 install -q -e .
 case "$1" in
 run) command=(dmoj) ;;
 cli) command=(dmoj-cli) ;;
-test) command=(python3 -- -m dmoj.testsuite testsuite) ;;
```
Xyene, line 14: Did you still mean to change this?

kiritofeng:
retest this please

kiritofeng:
The current code raises one warning per duplicate found (so if the code found `p1, p1, p2, p2, p2` for example, it would raise 1 warning for `p1` and 2 warnings for `p2`), along with the directories. Should I change this behaviour to only report each problem once / as one warning?

kiritofeng:
retest this please

kiritofeng:
As an example of what I mean by there being multiple warning, you'd have something that looks like
```
Skipped self-tests

Running local judge...

Warning: duplicate p1 found at /tmp/problems/b/p1, ignoring in favour of /tmp/problems/c/p1
Warning: duplicate p1 found at /tmp/problems/a/p1, ignoring in favour of /tmp/problems/c/p1
p1    
```
for config
```
problem_storage_globs:
- /tmp/problems/**/
```
and
```
/tmp/problems λ tree
.
├── a
│   └── p1
│       └── init.yml
├── b
│   └── p1
│       └── init.yml
└── c
    └── p1
        └── init.yml

7 directories, 3 files
```

Xyene (approved):
LGTM, thanks!

PR #1104, opened 2023-01-29
Fix bug where `get_supported_problems_and_mtimes` won't recursively scan

kiritofeng:
Also added a test.

Additional question: should we return a problem multiple times if it has different locations?


Xyene:
> Additional question: should we return a problem multiple times if it has different locations?

No, we should return the first one, and maybe print a warning about the second, but not return it.

Xyene (approved):
LGTM, thanks!

PR #1106, opened 2023-01-29
Declare types for `Result` and `CheckerResult`

kiritofeng:
Fixes a bug where interactive problems will crash upon grading a Java submission that returns `IR` due to `result.proc_output` being the empty string `''` by default. `result.proc_output` should now always be a bytestring.


Xyene (approved):
LGTM, thanks!

PR #1107, opened 2023-01-30
Detect JVM initialization errors

kiritofeng:
Changes:
1. print error to `stderr` instead of `stdout`, so we can detect this error within interactive graders
2. add more test cases, because JAVA8 has a slightly different error message, and there is a unique message at 2MB because god is dead and Java killed them


kiritofeng:
retest this please

Xyene (approved):
LGTM, thanks!

PR #1109, opened 2023-01-31
executors: add version checks to self-tests

### LINKED ISSUE #1108: CPP03 does not use C++03 dialect
`std` is not set to `c++03` in `CPP03.py`: https://github.com/DMOJ/judge-server/blob/96b2fffea81d84dd0f6c7f6a3f0d12a5900fb562/dmoj/executors/CPP03.py#L8

Hence CPP03 would actually use the default C++ dialect, which is C++17 for GCC 12. This is confirmed by compiling the test program for CPP17:
- C++03: https://dmoj.ca/submission/5267665 (compile ok)
- C++14: https://dmoj.ca/submission/5267672 (compile error)

kiritofeng:
Fixes #1108.

### DIFF
--- a/dmoj/executors/CPP03.py
+++ b/dmoj/executors/CPP03.py
@@ -8,10 +8,12 @@ class Executor(GCCExecutor):
     test_program = """
 #include <iostream>
 
+#if __cplusplus >= 199711 && __cplusplus < 201103
 int main() {
     std::cout << std::cin.rdbuf();
     return 0;
 }
+#endif
 """
 
     def get_flags(self):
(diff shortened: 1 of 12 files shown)


dmoj/executors/CPP03.py
```diff
@@ -1,20 +1,20 @@
-from typing import Optional
-
 from .gcc_executor import GCCExecutor
 
 
 class Executor(GCCExecutor):
```
Xyene, line 4: The fact that the C++03 executor is used as a "cpp_executor" of sorts is pretty yucky; we should just make a dedicated one at some point.

Xyene (approved):
LGTM, thanks!

PR #1110, opened 2023-02-23
Exec final command of docker entry file

Riolku:
This makes `docker stop` graceful, since otherwise, `bash` receives SIGTERM, and ignores it. With this setup, `runuser` receives SIGTERM.


Xyene (approved):
LGTM, thanks!

PR #1117, opened 2023-03-06
Use a real init process inside docker instead of runuser

quantum5:
`runuser` doesn't do the job of an `init` process such as reaping zombies.

### DIFF
--- a/.docker/tier1/Dockerfile
+++ b/.docker/tier1/Dockerfile
@@ -11,4 +11,4 @@ RUN mkdir /judge /problems && cd /judge && \
 	runuser -u judge -w PATH -- /env/bin/dmoj-autoconf -V > /judge-runtime-paths.yml && \
 	echo '  crt_x86_in_lib32: true' >> /judge-runtime-paths.yml
 
-ENTRYPOINT ["/judge/.docker/entry"]
+ENTRYPOINT ["/usr/bin/tini", "--", "/judge/.docker/entry"]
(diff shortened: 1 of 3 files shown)


Riolku:
I missed the discussions out of band, what's the concern with reaping zombies?

quantum5:
@Riolku go read https://stackoverflow.com/questions/49162358/docker-init-zombies-why-does-it-matter

.docker/tier1/Dockerfile
```diff
@@ -11,4 +11,4 @@ RUN mkdir /judge /problems && cd /judge && \
 	runuser -u judge -w PATH -- /env/bin/dmoj-autoconf -V > /judge-runtime-paths.yml && \
 	echo '  crt_x86_in_lib32: true' >> /judge-runtime-paths.yml
 
-ENTRYPOINT ["/judge/.docker/entry"]
+ENTRYPOINT ["/usr/bin/tini", "--", "/judge/.docker/entry"]
```
Riolku, line 14: What does `--` add?

Riolku: Like, specifically here: I know what it does in general.

Riolku: Oh it's for arguments that get passed on the command line on docker run.

Riolku (approved):

kiritofeng (approved):
LGTM

PR #1124, opened 2023-03-16
Use setpriv instead of runuser to avoid extra process

quantum5:
This prevents `runuser` from displaying the judge credentials to everyone on the same machine.


Xyene (approved):
LGTM

PR #1126, opened 2023-03-17
Add `locate` command

kiritofeng:
* Add `locate` command: output is
```
Skipped self-tests

Running local judge...

/tmp/problems/a
/tmp/problems/b
c is not a valid problem, skipping
```
for file structure
```
/tmp/problems » tree
.
├── a
│   └── init.yml
└── b
    └── init.yml

3 directories, 2 files
```
* Add typing annotations to `dmoj/judgeenv.py`
* Make `get_problem_root` return `None` if the problem does not exist
* Fix typo exposed in `dmoj/commands/resubmit.py` exposed by typing `dmoj/judgeenv.py`
* Fix typing annotation in `dmoj/packet.py`: `os.path.getmtime()` returns a `float` not `int`.

### DIFF
--- a/dmoj/tests/test_globs.py
+++ b/dmoj/tests/test_globs.py
@@ -89,7 +89,7 @@ class TestConfigGlobs(unittest.TestCase):
             ex_cases = ['p2', 'p3', 'doesnotexist']
 
             for problem in ex_cases:
-                self.assertRaises(KeyError, judgeenv.get_problem_root, problem)
+                self.assertIsNone(judgeenv.get_problem_root(problem))
 
     def test_supported_problems(self):
         with self.mock_problem_roots:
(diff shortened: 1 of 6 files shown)


dmoj/tests/test_globs.py
```diff
@@ -89,7 +89,7 @@ def test_problem_roots(self):
             ex_cases = ['p2', 'p3', 'doesnotexist']
 
             for problem in ex_cases:
-                self.assertRaises(KeyError, judgeenv.get_problem_root, problem)
+                self.assertIs(None, judgeenv.get_problem_root(problem))
```
quantum5, line 92:
```suggestion
                self.assertIsNone(judgeenv.get_problem_root(problem))
```

quantum5 (commented):
Seems mostly legit.

quantum5 (approved):

Issue #1129, opened 2023-04-06
Latest watchdog would trigger on_opened events


cuom1999:
Latest commit from watchdog: https://github.com/gorakhargosh/watchdog/commit/f3dfc4be0c6ec36c2db6b4afa075004e71adf87e

It means `SendProblemsHandler.on_any_event` in `monitor.py` would also trigger `on_opened`. Judges need to trigger this to read test data and init file. This leads to `supported-problems` being sent multiple times on every submission. Suppose there are n judges, bridged will receive at least 2 * n  `supported-problems` packets on each submission, which leads to 2n SQL update queries.

One simple fix is to add this line in `SendProblemsHandler.on_any_event`.
```python
def on_any_event(self, event):
        if event.event_type == "opened":
            return
        if self.callback is not None:
            self.callback()
        if self.refresher is not None:
            self.refresher.refresh()
```

Xyene:
That sounds reasonable, but I think rather than comparing against `opened` and bailing, we should compare against the set of events we do care about (in case `watchdog` adds more events in the future)

Would you be willing to submit a patch?

Xyene:
Fixed in https://github.com/DMOJ/judge-server/commit/17b0bcf3a0a79045e8dedfcbf2b47f4e2e724f93.

PR #1130, opened 2023-04-09
Add strict set of events to monitor

cuom1999:
This is to address watchdog's latest changes of adding `opened` and `closed` event_type.

https://github.com/DMOJ/judge-server/issues/1129#issuecomment-1501174321

### DIFF
--- a/dmoj/monitor.py
+++ b/dmoj/monitor.py
@@ -10,7 +10,13 @@ from dmoj.utils.glob_ext import find_glob_root
 
 try:
     from watchdog.observers import Observer
-    from watchdog.events import FileSystemEventHandler
+    from watchdog.events import (
+        FileSystemEventHandler,
+        EVENT_TYPE_MOVED,
+        EVENT_TYPE_DELETED,
+        EVENT_TYPE_MODIFIED,
+        EVENT_TYPE_CREATED,
+    )
 
     has_watchdog_installed = True
 except ImportError:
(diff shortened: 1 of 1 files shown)


dmoj/monitor.py
```diff
@@ -57,6 +57,14 @@ def __init__(self, refresher=None):
         self.callback = None
 
     def on_any_event(self, event):
+        ALLOWED_EVENT_TYPES = [
```
Xyene, line 60: Small nit: could we move this to inside `class SendProblemsHandler`, and have it be a tuple to indicate its immutability? Otherwise LGTM.

quantum5: Instead of tuple can we have a frozenset?

quantum5: On second thought `tuple` is fine given that it's just 4 elements.

dmoj/monitor.py
```diff
@@ -52,11 +52,20 @@ def run(self):
 
 
 class SendProblemsHandler(FileSystemEventHandler):
+    ALLOWED_EVENT_TYPES = (
+        'modified',
```
kiritofeng, line 56: Can you use the named constants https://github.com/gorakhargosh/watchdog/blob/f3dfc4be0c6ec36c2db6b4afa075004e71adf87e/src/watchdog/events.py#L103-L108 instead of hardcoding the strings?

kiritofeng (changes requested):

kiritofeng (approved):
LGTM, thanks!

PR #1147, opened 2023-12-25
executors: fully type every single executor

kiritofeng:
Supersedes #951.

### DIFF
--- a/dmoj/executors/base_executor.py
+++ b/dmoj/executors/base_executor.py
@@ -165,6 +165,9 @@ class BaseExecutor(metaclass=ExecutorMeta):
                 if exc.errno != errno.ENOENT:
                     raise
 
+    def create_files(self, problem_id: str, source_code: bytes, *args, **kwargs) -> None:
+        raise NotImplementedError()
+
     def __del__(self) -> None:
         self.cleanup()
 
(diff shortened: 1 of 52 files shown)


dmoj/executors/base_executor.py
```diff
@@ -157,6 +165,9 @@ def cleanup(self) -> None:
                 if exc.errno != errno.ENOENT:
                     raise
 
+    def create_files(self, problem_id: str, source_code: bytes, *args, **kwargs):
```
quantum5, line 168:
```suggestion
    def create_files(self, problem_id: str, source_code: bytes, *args, **kwargs) -> None:
```

dmoj/executors/java_executor.py
```diff
@@ -268,7 +271,7 @@ def get_compile_args(self) -> List[str]:
         compiler = self.get_compiler()
         assert compiler is not None
         assert self._code is not None
-        return [compiler, '-Xlint', '-encoding', 'UTF-8', self._code]
+        return [compiler, '-encoding', 'UTF-8', self._code]
```
quantum5, line 274: Why did we kill `-Xlint`?

quantum5 (approved):

dmoj/executors/java_executor.py
kiritofeng: Neither `JAVA.py` nor `JAVA8.py` use `-Xlint` from what I can tell.

PR #1154, opened 2023-12-27
testsuite: add rust generator test

Riolku:
This would have caught a failure on #1148.

### DIFF
--- a/testsuite/rust_generator/tests/AC/test.yml
+++ b/testsuite/rust_generator/tests/AC/test.yml
@@ -1,0 +1,5 @@
+language: PY3
+time: 2
+memory: 65536
+source: sol.py
+expect: AC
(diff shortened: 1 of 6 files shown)


testsuite/rust_generator/tests/AC/test.yml
```diff
@@ -0,0 +1,5 @@
+language: PY3
+time: 2
+memory: 65536
+source: sol.py
+expect: [AC, AC, AC]
```
kiritofeng, line 5:
```suggestion
expect: AC
```

kiritofeng (commented):
Otherwise LGTM, thanks!

testsuite/rust_generator/tests/AC/test.yml
Riolku: Also updated the test I copied this from, lol.

Riolku:
Also, I did in fact test this against #1148.

kiritofeng (approved):
LGTM, thanks!

PR #1155, opened 2023-12-31
executors: rework GCC/Clang executors

quantum5:
Changes:

* Make -std flags explicit everywhere.
* Clang no longer inherits from GCC.
* All C and C-like executors now inherit from CLikeExecutor.
* Add -std=f95 for F95 executor.
* CLANGX now correctly passes -std=c++14.


Xyene (approved):
LGTM

PR #1156, opened 2024-01-02
Various fixes to restore FreeBSD support

quantum5:


testsuite/rust_generator/tests/AC/test.yml
```diff
@@ -1,4 +1,5 @@
 language: PY3
+executors: [RUST]
```
kiritofeng, line 2: Can we rename this to `required_auxiliary_executors` so it's clear what this is for?

dmoj/utils/cpp_demangle.py
```diff
@@ -1,13 +1,23 @@
...
+
+libstdcxx: Any = None if libstdcxx_path is None else CDLL(libstdcxx_path)
+libcxx: Any = None if libcxx_path is None else CDLL(libcxx_path)
 
 # Signature: char *__cxa_demangle(const char *mangled_name, char *output_buffer, size_t *length, int *status)
-__cxa_demangle = libstdcxx.__cxa_demangle
+try:
+    __cxa_demangle = libstdcxx.__cxa_demangle
+except AttributeError:
+    __cxa_demangle = libcxx.__cxa_demangle
```
kiritofeng, line 19: Do we need to do anything here if this also fails? (e.g. `libcxx` is `None`?)

kiritofeng (commented):
Otherwise LGTM

dmoj/utils/cpp_demangle.py
quantum5: You are screwed then, it'll crash.

testsuite/rust_generator/tests/AC/test.yml
quantum5: Renamed.

kiritofeng (approved):
LGTM

PR #1160, opened 2024-01-08
Run initialize in CI tests

int-y1:
Add `load_executors()` so that `initialize` gets called in CI tests.

### DIFF
--- a/dmoj/citest.py
+++ b/dmoj/citest.py
@@ -1,13 +1,12 @@
 import glob
 import os
 import traceback
-from importlib import import_module
 
 import yaml
 
 from dmoj import judgeenv
 from dmoj.contrib import load_contrib_modules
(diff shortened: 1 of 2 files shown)


quantum5:
ok to test

dmoj/citest.py
```diff
@@ -17,8 +16,10 @@ def ci_test(executors_to_test, overrides, allow_fail=frozenset()):
     failed = False
     failed_executors = []
 
+    load_executors()
```
quantum5, line 19: Are you calling this for the side-effect of loading all executors into `sys.modules`? Feels kinda weird since it prints `Self-testing executors` and does a bunch of other things. Might be better to refactor the loading logic in `load_modules` into a function and call that, then don't load these modules.

PR #1177, opened 2024-06-02
tests/check_memory_limit: increase memory usage

Riolku:
FreeBSD memory accounting is not very accurate - increasing the memory usage should make this test more robust.


Xyene (approved):
LGTM, thanks!

Issue #1182, opened 2024-07-03
Build error (distutils, CCompiler) when building tier1 judge


dinoilic:
Hi all,

Not sure if this is the right repo to submit this issue, but I'm following the instructions on setting up a docker version of a judge, and after running `make judge-tier1` it runs for a while and ends up with the following:

```
19.89 Successfully built dmoj
19.90 Installing collected packages: pylru, watchdog, termcolor, setproctitle, pyyaml, pygments, dmoj
21.06 Successfully installed dmoj-4.1.0 pygments-2.18.0 pylru-1.2.1 pyyaml-6.0.1 setproctitle-1.3.3 termcolor-2.4.0 watchdog-4.0.1
21.21 Traceback (most recent call last):
21.21   File "/judge/setup.py", line 6, in <module>
21.21     from distutils.ccompiler import CCompiler
21.21 ModuleNotFoundError: No module named 'distutils'
------
Dockerfile:4
--------------------
   3 |     ARG TAG=master
   4 | >>> RUN mkdir /judge /problems && cd /judge && \
   5 | >>> 	curl -L https://github.com/DMOJ/judge-server/archive/"${TAG}".tar.gz | tar -xz --strip-components=1 && \
   6 | >>>     python3 -m venv --prompt=DMOJ /env && \
   7 | >>> 	/env/bin/pip3 install cython && \
   8 | >>> 	/env/bin/pip3 install -e . && \
   9 | >>> 	/env/bin/python3 setup.py develop && \
  10 | >>> 	HOME=~judge . ~judge/.profile && \
  11 | >>> 	runuser -u judge -w PATH -- /env/bin/dmoj-autoconf -V > /judge-runtime-paths.yml && \
  12 | >>> 	echo '  crt_x86_in_lib32: true' >> /judge-runtime-paths.yml
  13 |
--------------------
[...]

PR #1184, opened 2024-07-11
Install Module distutils

Elfsong:
Build error (distutils, CCompiler) when building tier1 judge 
#1182


dinoilic:
Thanks for fixing this! I read that this would indeed solve our issue, but I don't know how this affects existing systems since I believe the build has been passing (or partially passing) a month ago.

Screenshot 2024-07-11 at 09 02 27

Anyways, I don't have the permissions to approve either, so I'll let the admins help out when they can :)

PR #1214, opened 2026-03-10
upgrade actions to python 3.11

int-y1:
currently limited to python 3.11, because one of the actions uses https://github.com/DMOJ/runtimes-python


int-y1:
github workers stopped supporting python 3.7, and upgrading to 3.11 will give us a few more years of support

PR #1216, opened 2026-03-11
executors: fix executor setup

int-y1:
* end support for `ICK`
* temporarily end support for `SCALA` (we can re-enable scala after upgrading from 2.11.12 (nov 2017))
* allow `DART` to call `execve` on `/opt/dart-sdk/bin/dartvm`
* allow `HASK` and `KOTLIN` to use `getresuid` and `getresgid`

### DIFF
--- a/dmoj/executors/DART.py
+++ b/dmoj/executors/DART.py
@@ -1,6 +1,7 @@
 from typing import List
 
 from dmoj.cptbox.filesystem_policies import ExactFile, RecursiveDir
+from dmoj.cptbox.isolate import DeniedSyscall, protection_fault
 from dmoj.executors.compiled_executor import CompiledExecutor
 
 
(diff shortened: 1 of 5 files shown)


dmoj/executors/DART.py
```diff
@@ -29,6 +30,19 @@ class Executor(CompiledExecutor):
...
     ]
 
+    def get_security(self, launch_kwargs=None, extra_fs=None):
+        from dmoj.cptbox.syscalls import sys_execve
+
+        sec = super().get_security(launch_kwargs=launch_kwargs, extra_fs=extra_fs)
+
+        def handle_execve(debugger) -> None:
+            path = debugger.readstr(debugger.uarg0)
+            if path != '/opt/dart-sdk/bin/dartvm':
```
kiritofeng: @quantum5 is there a way to avoid hardcoding this path?

PR #1227, opened 2026-03-29
executors: change terminal width to 80

kiritofeng:
Scala uses the terminal width to decide how long to make the lines it uses in its error messages.
Currently, the terminal is set to a width of 1024, leading to silly behaviour like this:
image

By changing the terminal width to 80, we get much more reasonable behaviour
```scala
-- [E053] Type Error: /tmp/tmpas35nd3u/helloworld.scala:3:19 -------------------
3 |    def main(args: String[Array]) = {
  |                   ^^^^^^^^^^^^^
  |                   String does not take type parameters
  |
  | longer explanation available when compiling with `-explain`
1 error found
```

### DIFF
--- a/dmoj/executors/compiled_executor.py
+++ b/dmoj/executors/compiled_executor.py
@@ -139,8 +139,8 @@ class CompiledExecutor(BaseExecutor, metaclass=_CompiledExecutorMeta):
         _master, _slave = pty.openpty()
 
         # Some runtimes helpfully try to word-wrap error messages by determining the width of the screen. Lie and say
-        # we're a 1024x1024 terminal, so they don't try wrapping to 1-column width.
-        fcntl.ioctl(_slave, termios.TIOCSWINSZ, struct.pack('HHHH', 1024, 1024, 0, 0))
+        # we're a 80x1024 terminal, so they don't try wrapping to 1-column width.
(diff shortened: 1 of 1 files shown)


dmoj/executors/compiled_executor.py
```diff
@@ -139,8 +139,8 @@ def create_compile_process(self, args: List[str]) -> TracedPopen:
         _master, _slave = pty.openpty()
 
         # Some runtimes helpfully try to word-wrap error messages by determining the width of the screen. Lie and say
-        # we're a 1024x1024 terminal, so they don't try wrapping to 1-column width.
-        fcntl.ioctl(_slave, termios.TIOCSWINSZ, struct.pack('HHHH', 1024, 1024, 0, 0))
+        # we're a 1024x80 terminal, so they don't try wrapping to 1-column width.
+        fcntl.ioctl(_slave, termios.TIOCSWINSZ, struct.pack('HHHH', 1024, 80, 0, 0))
```
quantum5: You mean 80x1024? Also why 80? Seems a bit narrow for the 21st century.

kiritofeng: Oops, I always get my matrix dimensions backwards...

As for why it's 80, I thought about making it 120, but this is too wide if I use DMOJ split-screen on my monitor (I expect that DMOJ + editor isn't too uncommon a setup, and it has the advantage of playing nice with phones).

PR #1228, opened 2026-03-29
C/++ executor updates

### LINKED ISSUE #1218: Add `is_siggradable` flag to executors instead of hardcoding
We should add an `is_siggradable` flag to the executors that can be signature graded, rather than hardcoding the list into the `SignatureGrader` class.

This also has the marginal benefit that any custom graders which override the signature grader do not need to be updated every time we add a new language.

kiritofeng:
A series of updates to C/C++/Clang++ executors.

1. Change the `__cpluscplus` comparison in the `CPP23` executor to equality: this way we don't get trolled when C++26 is added.
2. Add new Clang++ executors for C++17, C++20, C++23
3. Add C23 executor
4. Rename the old `CLANGX` executor to `CLANGX14` to match the other executors.
5. Add `is_signature_graded` flag, so we don't have to fix the hard coded list in `signature.py` every time; fixes #1218.


.freebsd.test.py
```diff
@@ -11,8 +11,10 @@
     'CPP11',
     'CPP14',
     'CPP17',
+    'CPP20',
     'CLANG',
-    'CLANGX',
+    'CLANGX14',
+    'CLANGX23',
```
quantum5: Guess FreeBSD doesn't need CLANGX17 or CLANGX20?

.freebsd.test.py
```diff
@@ -11,8 +11,10 @@
     'CPP11',
     'CPP14',
     'CPP17',
+    'CPP20',
     'CLANG',
-    'CLANGX',
+    'CLANGX14',
```
quantum5: Is this too long for the site? Language.key is capped at 6 characters.

.freebsd.test.py
kiritofeng: I can add them both; for future reference, we don't test `CPP23` because the `g++` on our FreeBSD box doesn't support C++23 yet.

.freebsd.test.py
kiritofeng: Ok this is a problem. What would you recommend we call the language? We could do `CLPP14` for CLang++14 for example?

quantum5: CLPP, CLX, CLXX, I don't really care.

kiritofeng: Fixed; they are now `CLPP<version>`

.freebsd.test.py
kiritofeng: Added.

quantum5 (approved):
Seems legit. I trust you to handle the site-side migration.

PR #1230, opened 2026-03-29
Implement Prometheus integration

quantum5:


dmoj/judge.py
```diff
@@ -201,6 +211,7 @@ def _ipc_grading_begin(self, _report, is_pretested: bool) -> None:
 
     def _ipc_grading_end(self, _report) -> None:
         self.packet_manager.grading_end_packet()
+        self.submission_graded += 1
```
kiritofeng: nit: Should we consider renaming this to `submissions_graded`?

kiritofeng (approved):
Minor nit but otherwise LGTM.

dmoj/judge.py
quantum5: Renamed.

PR #1235, opened 2026-05-03
Fix FreeBSD syscall list breakage and ensure it can never happen again

quantum5:
This PR loads the FreeBSD syscall list from GitHub instead of dealing with FreeBSD's bot protection. Also rewrote the script to use requests so that it will appropriately crash instead of sending empty files.


kiritofeng (approved):
LGTM

PR #1238, opened 2026-06-04
feat: support case_batch_dependencies when using regex test case format

### LINKED ISSUE #1232: Allow batch dependencies when using regexes
We should allow for the use of a `case_batch_dependencies` key when using regexes to specify problem configurations.
This would eliminate one of the final points of friction for using regexes to configure problems.

Rishiii57:
Fixes #1232 

Adds support for case_batch_dependencies when using regex test case format specifiers.

The key accepts an array of arrays, where each index corresponds to a batch:

test_cases:
  case_batch_dependencies: [[], [1], [1], [], [1,2,3,4]]

Changes:
- Added case_batch_dependencies parameter to _match_test_cases
- Read and unwrap case_batch_dependencies from test_cases config node in _resolve_test_cases
- Inject dependencies into batch dicts using array indexing


kiritofeng:
Can you add a test for this to the testsuite? You can base it off of https://github.com/DMOJ/judge-server/blob/master/testsuite/batch_dependencies/init.yml.

Rishiii57:
Hey I've looked at the testsuite structure. Could you give me some guidance on how to add a regex-based test there? I'm on a Mac so I can't run the full judge environment locally to verify, since cptbox requires Linux.

kiritofeng:
You probably want to make `{1,2,3,4}.txt` (containing 1, 2, 3, 1) and then write `init.yml`
```yml
test_cases:
  input_format: (?P<case>\d)\.txt
  output_format: (?P<case>\d)\.txt
  case_batch_dependencies: [[], [], [1, 2], []]
```
and the rest can presumably be copied over without modification.

Rishiii57:
Yeah thats what I thought.
Done:)

Rishiii57:
Hey @kiritofeng I tried doing what you said but seems like some checks are failing. Could you please guide me what to do?

kiritofeng:
You also need to handle the case where the test case is not in a batch: https://github.com/DMOJ/judge-server/blob/master/dmoj/problem.py#L159

I also think that calling them `case_dependencies` instead of `case_batch_dependencies` might be a bit better

Rishiii57:
Hey @kiritofeng , I have updated the PR based on your suggestions :)
please have a look

kiritofeng:
The code is failing because the dependencies are not being correctly assigned; you should output the generated config and see why it's failing.

kiritofeng:
It's failing because the cases aren't being generated as batches but rather as individual test cases (and `dependencies` is literally `batch_dependencies`, i.e. it only works on batches), so you should probably rename it back to `case_batch_dependencies` and change the config so it generates batches instead of individual test cases.

PR #1239, opened 2026-06-07
Fix `feedback` flag

### LINKED ISSUE #1165: `feedback` flag shouldn't apply for IE
If the `feedback` flag of a custom checker (e.g. of type `coci` or `testlib`) is set to `false`, then feedback won't be included in the error message in case of an IE. I don't see any cases where it is desirable not to have the additional information for debugging, so the `feedback` flag should only apply for feedback to the user.

kiritofeng:
This PR separates `feedback` into an actual `feedback` parameter and a `show_feedback` boolean.
This way we always have `feedback` available for parsing when raising an internal error, but can also choose not to show it to the contestant.

Fixes #1165

### DIFF
--- a/dmoj/checkers/bridged.py
+++ b/dmoj/checkers/bridged.py
@@ -43,7 +43,7 @@ def check(
     executor = get_executor(problem_id, files, flags, lang, compiler_time_limit)
 
     if type not in contrib_modules:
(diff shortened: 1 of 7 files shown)


quantum5 (approved):
Sure, I guess? Doesn't look particularly broken.

PR #1240, opened 2026-06-07
Simplify generator specification

### LINKED ISSUE #1202: Simplify generator spec
The current generator spec consists of several hacks to handle previous unintended use cases (primarily that if the generator crashed, the input/output files would be passed directly to the program).

This has led to an opaque generator spec that is not documented and has several counterintuitive behaviours.
For example, specifying `input_file` and no `output_file` will run the generator with the input file passed to the generator as file output, but if an output file is then specified, the generator will be totally bypassed.

I propose rewriting the spec for generator behaviour to the following:
- If `generator_args` are specified, the generator is run with arguments (no and file input). Including `input_file` or `generator_input_file`  will result in an `InvalidInitException`.
  - If `output_file` is included, it will be passed as the judge output.
- Otherwise:
  - If `generator_input_file` is specified, the generator is fed the input file. Including `input_file` will result in an `InvalidInitException`.
    - If `output_file` is included, it will be passed as the judge output.
  - Otherwise:
    - If `input_file` is specified, the generator is ignored for that test case, and the files are passed directly to the submission
    - Otherwise, the generator is run, and `output_file` is passed as judge output if it exists.

This should hopefully result in more intuitive and well-documented behaviour.

kiritofeng:
Fixes #1202.

Previously whether or not a generator was run depended on checking if at least one of `in` or `out` were missing.
This could lead to counterintuitive situations where you would have to specify a dummy output file to prevent the generator from treating the input file as its own input.

This PR proposes the new specification for generators:
1. Check if a test case has an `in` node specified, i.e. an explicit input file. If so, this is treated as the input file. The generator is not run. Whether or not an output file is specified is no longer relevant to the check.
2. If not, the generator is run. `generator_args`, if present, are passed as arguments to the generator, and `generator_in`, if present, are passed to the generator as the input file.

When parsing `init.yml`, there is also a check performed that the method of supplying the input is unambiguous, i.e. a node cannot have both `in` and `generator_args`/`generator_in` specified.

There are 16 problems on DMOJ whose generators rely on reading input from a file. These would need to have their `init.yml` configs rewritten to replace `in` with `generator_in`.


dmoj/problem.py
```diff
@@ -443,7 +443,7 @@ def _run_generator(self, gen: Union[str, ConfigNode], args: Optional[Iterable[st
         )
 
         try:
-            input = self.problem.problem_data[self.config['in']] if self.config['in'] else None
+            input = self.problem.problem_data[self.config['generator_in']] if self.config['generator_in'] else None
```
quantum5: Can we just like... save `self.config['generator_in']` to a variable or something? I don't like the way this looks.

dmoj/problem.py
```diff
@@ -466,9 +466,19 @@ def input_data_io(self) -> MmapableIO:
...
+        # we follow the following logic:
+        # check if the case has an input file specified
+        # - if yes, use that file as input, and make sure `generator_args` and `generator_in` are not specified
+        # - if no, run the generator and pass `generator_args` and `generator_in`
+        #   it is ok if both are specified
+        run_generator = bool(gen)
+        if run_generator and self.config['in']:
+            run_generator = False
+            if self.config['generator_args'] or self.config['generator_in']:
+                raise InvalidInitException('Input for test case ambiguous! Make up your mind!')
```
quantum5, line 480: Reminds me of that error message in bird...

quantum5 (approved):
Seems fine, minor style nit.

PR #1241, opened 2026-06-08
Fix language autodetection

### LINKED ISSUE #967: Auto-detect generator language when not explicitly specified

### LINKED ISSUE #1233: Fix Clang++ Signature Grading
Currently it fails with compile error.

```
clang++: warning: treating 'c-header' input as 'c++-header' when in C++ mode, this behavior is deprecated [-Wdeprecated]
clang++: error: cannot specify -o when generating multiple output files
```

The solution might be to disable signature grading in Clang/++.

kiritofeng:
This PR fixes language auto-detection, supporting sorting by version using the `ext_priority` parameter set in the executor.

This, however, raises the possibility of generators/interactors/etc also being compiled with `clang`/`clang++`. Thus it was necessary to also fix multifile support for those languages.

Fixes #967; fixes #1233.

### DIFF
--- a/dmoj/executors/C.py
+++ b/dmoj/executors/C.py
@@ -4,6 +4,7 @@ from dmoj.executors.c_like_executor import CExecutor, GCCMixin
 class Executor(GCCMixin, CExecutor):
     command = 'gcc'
     std = 'c99'
+    ext_priority = 1
 
     test_program = """
 #include <stdio.h>
--- a/dmoj/executors/__init__.py
+++ b/dmoj/executors/__init__.py
@@ -22,8 +22,8 @@ def by_ext(ext: str) -> Any:
         if name.lower() == ext:
             return executor
 
-    for executor in sorted(executors.values(), key=lambda executor: executor.Executor.name):
-        if executor.Executor.ext == ext:
+    for executor in sorted(executors.values(), key=lambda executor: executor.Executor.ext_priority, reverse=True):
+        if ext in executor.Executor.get_valid_exts():
             return executor
 
     raise KeyError('no executor for extension "%s"' % ext)
--- a/dmoj/executors/base_executor.py
+++ b/dmoj/executors/base_executor.py
@@ -6,7 +6,7 @@ import subprocess
 import sys
 import tempfile
 import traceback
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Type, Union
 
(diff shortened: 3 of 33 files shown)


dmoj/executors/__init__.py
```diff
@@ -22,8 +22,8 @@ def by_ext(ext: str) -> Any:
         if name.lower() == ext:
             return executor
 
-    for executor in sorted(executors.values(), key=lambda executor: executor.Executor.name):
-        if executor.Executor.ext == ext:
+    for executor in sorted(executors.values(), key=lambda executor: executor.Executor.ext_priority, reverse=True):
```
quantum5, line 25: Can we specify a fixed order? Like lexicographically increasing by executor name? I don't want any flaky behaviour and dict value order isn't guaranteed...

dmoj/executors/base_executor.py
```diff
@@ -483,3 +484,12 @@ def get_find_first_mapping(cls) -> Optional[Dict[str, List[str]]]:
     @classmethod
     def autoconfig(cls) -> AutoConfigOutput:
         return cls.autoconfig_find_first(cls.get_find_first_mapping())
+
+    @classmethod
+    def get_valid_exts(cls) -> Sequence[str]:
```
quantum5, line 489: Is there a reason why this is a method and not just a class variable?

dmoj/executors/C.py
```diff
@@ -4,6 +4,7 @@
 class Executor(GCCMixin, CExecutor):
     command = 'gcc'
     std = 'c99'
+    ext_priority = 1
```
quantum5, line 7: Can we use a smarter numbering scheme? Like 111 for Clang C11, 10099 for GCC C99, 10111 for C11, 11114 for C++14, etc.? This way, we don't need to shuffle all the numbers to add a new runtime...

dmoj/executors/base_executor.py
```diff
@@ -483,3 +484,12 @@ def get_find_first_mapping(cls) -> Optional[Dict[str, List[str]]]:
...
+        # This exists because sometimes we want to recognize more extensions than just `ext` for auxiliary files
+        return (cls.ext,)
+
+    @classmethod
+    def supports_multifile(cls) -> bool:
```
quantum5, line 494: Similarly, class variable.

PR #1244, opened 2026-06-15
 ci: use GitHub's native arm and more Python versions

quantum5:
Also do some fixes for arm-specific prctls.

### DIFF
--- a/dmoj/cptbox/isolate.py
+++ b/dmoj/cptbox/isolate.py
@@ -407,6 +407,7 @@ class IsolateTracer(dict):
         PR_GET_NAME = 16
         PR_CAPBSET_READ = 23
         PR_SET_THP_DISABLE = 41
+        PR_SVE_SET_VL = 50
         PR_SVE_GET_VL = 51
         PR_SET_VMA = 0x53564D41  # Used on Android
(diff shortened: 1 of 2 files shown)


dmoj/cptbox/isolate.py
```diff
@@ -407,6 +407,8 @@ def handle_prctl(self, debugger: Debugger) -> None:
         PR_GET_NAME = 16
         PR_CAPBSET_READ = 23
         PR_SET_THP_DISABLE = 41
+        PR_SVE_GET_VL = 51
```
kiritofeng: The order of this and the next line should be swapped to be consistent with the rest of the code (they're ordered numerically)

dmoj/cptbox/isolate.py
```diff
@@ -415,6 +417,8 @@ def handle_prctl(self, debugger: Debugger) -> None:
             PR_CAPBSET_READ,
             PR_SET_THP_DISABLE,
             PR_SET_VMA,
+            PR_SVE_GET_VL,
```
kiritofeng: Same here.

kiritofeng (approved):
Minor nit but otherwise LGTM

PR #1246, opened 2026-06-15
executors/ZIG: update to support 0.16.0

int-y1:
newer version of #1221 because zig refactored its library again

i have no idea what `&.{&buf}` means

```
0.527 Testing executors...
0.562 Testing ZIG:    Using /opt/zig/zig
20.89   zig: 0.16.0
20.89
```


kiritofeng (approved):
I believe you.

PR #1251, opened 2026-08-07
executors/MONO*: fix compile error involving TERM

int-y1:
apparently, `env['TERM'] = 'xterm'` causes MONOCS/MONOFS to compile error

the cause is probably `xterm` getting too bloated recently

before:
```
0.536 Testing MONOCS: Failed self-test
0.632   Attempted:
0.632     mono: /usr/bin/mono
0.633     mono-csc: /usr/bin/mono-csc
0.633   Errors:
0.633     Traceback (most recent call last):
...
0.633     dmoj.error.CompileError:
0.633     Unhandled Exception:
0.633     System.TypeInitializationException: The type initializer for 'System.Console' threw an exception. ---> System.TypeInitializationException: The type initializer for 'System.ConsoleDriver' threw an exception. ---> System.Exception: File must be smaller than 4K
```

after:
```
0.672 Testing MONOCS: Success
1.525   csc: 6.12.0.200
1.536   mono: 6.12.0.200
```

### DIFF
--- a/dmoj/executors/mono_executor.py
+++ b/dmoj/executors/mono_executor.py
@@ -45,6 +45,11 @@ class MonoExecutor(CompiledExecutor):
         ('fork', ACCESS_EAGAIN),
     ]
 
+    def get_compile_env(self) -> Dict[str, str]:
+        env = super().get_compile_env() or os.environ.copy()
+        env['TERM'] = 'dumb'
+        return env
+
     def get_env(self) -> Dict[str, str]:
         env = super().get_env()
         # Disable Mono's usage of /dev/shm, so we don't have to deal with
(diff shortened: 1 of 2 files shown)


dmoj/executors/mono_executor.py
```diff
@@ -45,6 +45,11 @@ class MonoExecutor(CompiledExecutor):
         ('fork', ACCESS_EAGAIN),
     ]
 
+    def get_compile_env(self) -> Dict[str, str]:
+        env = super().get_compile_env() or os.environ.copy()
+        env['TERM'] = ''
```
quantum5: I wonder if we should do `TERM=dumb` instead.

quantum5 (approved):
ok to test

dmoj/executors/mono_executor.py
int-y1: `TERM=dumb` seems reasonable, and i made the change

PR #1253, opened 2026-08-07
executors/LEAN4: reduce memory use and increase grace

int-y1:
before:
```
0.521 Testing LEAN4:  Failed self-test
1.111   Attempted:
1.111     lean: /opt/lean/bin/lean
1.112   Errors:
1.112     Traceback (most recent call last):
...
1.112     dmoj.error.CompileError: uncaught exception: failed to locate application
```

after the `isolate.py` fix:
```
0.507 Testing LEAN4:  libc++abi: terminating due to uncaught exception of type lean::exception: failed to create thread                                 
1.414 Failed self-test
1.414   Attempted:
1.414     lean: /opt/lean/bin/lean
1.415   Errors:
1.415     Got unexpected stdout output:
```

after the `LEAN4.py` fix:
```
0.518 Testing LEAN4:  Using /opt/lean/bin/lean
1.563   lean: 4.32.2
```

`LEAN_MAIN_USE_THREAD` and `MIMALLOC_ARENA_RESERVE` reduce the grace needed. `LEAN_NUM_THREADS` reduces the grace for programs that spawn threads (maybe?) (i can't imagine users using multithreading).

the grace is 1gb for the main thread, plus 64mb for mimalloc.

### DIFF
--- a/dmoj/cptbox/isolate.py
+++ b/dmoj/cptbox/isolate.py
@@ -357,9 +357,12 @@ class IsolateTracer(dict):
             file = os.path.join(f'/proc/{debugger.tid}', os.path.relpath(file, '/proc/self'))
             projected = '/' + os.path.normpath(file).lstrip('/')
         elif normalized.startswith(f'/proc/{debugger.tid}/'):
-            # If the child process uses /proc/getpid()/foo, set the normalized path to be /proc/self/foo.
+            # If the child process uses /proc/[tid]/foo, set the normalized path to be /proc/self/foo.
             # Access rules can more easily check /proc/self.
             normalized = os.path.join('/proc/self', os.path.relpath(file, f'/proc/{debugger.tid}'))
+        elif normalized.startswith(f'/proc/{debugger.pid}/'):
+            # Normalize /proc/[pid]/foo to /proc/self/foo. LEAN4 uses this.
+            normalized = os.path.join('/proc/self', os.path.relpath(file, f'/proc/{debugger.pid}'))
         real = os.path.realpath(file)
 
         try:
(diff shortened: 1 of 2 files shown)


kiritofeng:
ok to test

dmoj/cptbox/isolate.py
```diff
@@ -360,6 +360,8 @@ def _access_check(self, debugger: Debugger, file: str, fs_jail: FilesystemPolicy
             # If the child process uses /proc/getpid()/foo, set the normalized path to be /proc/self/foo.
```
kiritofeng: This shouldn't be `/proc/getpid()/foo` but `/prod/gettid()/foo`

dmoj/cptbox/isolate.py
```diff
@@ -360,6 +360,8 @@ def _access_check(self, debugger: Debugger, file: str, fs_jail: FilesystemPolicy
...
             # Access rules can more easily check /proc/self.
             normalized = os.path.join('/proc/self', os.path.relpath(file, f'/proc/{debugger.tid}'))
+        elif normalized.startswith(f'/proc/{debugger.pid}/'):
```
kiritofeng, line 363: Probably want to refactor the comment for the above `tid()` case here (and add the reason we need both checks is because of multi-threaded executors like Lean and these agree for non-multi-threaded)

dmoj/executors/LEAN4.py
```diff
@@ -12,6 +14,8 @@ def main : IO Unit := do
 """
     nproc = -1
     syscalls = ['eventfd2']
+    address_grace = 1088 * 1024
+    data_grace = 1088 * 1024
```
kiritofeng, line 18: I'm assuming these numbers were tested

kiritofeng (approved):
Mostly LGTM; can you just refactor the comments in `isolate.py` a bit?

dmoj/executors/LEAN4.py
int-y1: they're tested, and lowering each one by 20% results in `lean::exception: failed to create thread`

